PR #9, opened 2025-10-05
Resolve testing Issues

riasc:
## Summary by CodeRabbit

* **Tests**
  * Enabled unified test discovery and execution so tests are built only when testing is turned on; test runs are more reliable and organized into clearer groups for improved reporting.

* **Chores**
  * CI workflow clarified: renamed a build step and added an optional debug check to inspect test configuration before running tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,23 +73,23 @@ jobs:
         run: |
           cmake --build ${{ steps.strings.outputs.build-output-dir }} --config ${{ matrix.build_type }}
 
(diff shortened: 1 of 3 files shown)


riasc:
Thanks for letting me know. Its kinda weird. I will revert this as this does not resolve this. Need to test this more. 

Yes, we should do that. I also added MVSC, but since removed it as this had a lot of issues with the paths.

cauliyang:
sounds great! I will continue to make some small changes and push a new pr.

riasc:
Nice. Thanks for your help

PR #13, opened 2025-10-06
Optimize vector operations with reserve() calls, std::string_view

cauliyang:
- using std::string_view
- Add reserve() in node constructor for keys and children vectors
- Add reserve() in calc_parent_key() to avoid reallocations
- All tests pass with no performance regression


riasc:
Didn't know about string_view.. good one.

PR #25, opened 2025-10-16
Feat: Use template to replace type register

cauliyang:
# What Changed

  Removed (Type Erasure Approach):
  - type_registry.hpp / type_registry.cpp - Global singleton registry
  - any_type.hpp - Type-erased data storage
  - key_old.hpp - Old key implementation using any_base*
  - type_registry-test.cpp - Tests for old system

  Added/Updated (Zero-Overhead Templates):
  - key.hpp - Modern template-based key with key<key_type, data_type = void>
  - grove.hpp - Now templated on both key_type and data_type
  - node.hpp - Now templated on both parameters
  - query_result.hpp - Now templated on both parameters
  - All test files updated to use new API

  # Key Improvements

  1. Zero Runtime Overhead: No virtual functions, no type erasure, no global state
  2. Compile-Time Type Safety: Uses C++20 concepts and requires clauses
  3. Better Performance: [[no_unique_address]] optimization, perfect forwarding
  4. Simpler API: Direct type access instead of type_registry::cast<T>()
  5. Extensible: Custom types via serialization_traits<T> specialization

  # Example Usage

  // Old way (removed):
  gdt::type_registry::register_type<int>();
  gdt::key<interval> key(intvl, 15);
  auto data = gdt::type_registry::cast<int>(key.get_data());

  // New way:
  gdt::key<interval, int> key(intvl, 15);  // Type specified at compile time
  int data = key.get_data();  // Direct access, type-safe

@riasc, we may need to conduct runtime benchmarking with existing tools again, and the current implementation should run faster in theory.

## Summary by CodeRabbit

* **New Features**
  * Keys and query results now support associated typed data; direct data access and presence checks added.
  * Grove structure accepts and propagates keyed data types.

* **Build Changes**
  * Added configurable BUILD_BENCHMARKS option and updated help text.

* **Bug Fixes / Chores**
  * Removed legacy runtime type-registration/any-type mechanisms (API simplified).

* **Tests**
  * Expanded and updated tests for key/query_result/grove with data-bearing keys and heterogeneous data support.

### DIFF
--- a/tests/structure/heterogeneous-grove-test.cpp
+++ b/tests/structure/heterogeneous-grove-test.cpp
@@ -1,0 +1,179 @@
+/*
+ * SPDX-License-Identifier: MIT
+ *
+ * Copyright (c) 2025 Richard A. Schäfer
+ *
+ * This file is part of genogrove and is licensed under the terms of the MIT license.
+ * See the LICENSE file in the root of the repository for more information.
+ */
+
+// Google Test
+#include <gtest/gtest.h>
+
+// Standard
+#include <variant>
+#include <string>
+
+// genogrove
+#include <genogrove/data_type/interval.hpp>
+#include <genogrove/structure/grove/grove.hpp>
+
+namespace gdt = genogrove::data_type;
+namespace gst = genogrove::structure;
+
+// Define variant type for testing
+using TestVariant = std::variant<std::monostate, int, double, std::string>;
+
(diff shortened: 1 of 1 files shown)


riasc:
Thanks for this.. its similar as I tried before. The thing was always that I needed to store different types in the tree, rather than always the same type. I need to check first if templating a std::variant works ok... as this is how I implemented it within the transcript discovery. But generally the usability with the type register is not the greatest.. :(

PR #302, opened 2026-04-08
Clear source stream state after inflate Z_STREAM_END

### LINKED ISSUE #301: inflate_streambuf sets failbit on underlying stream after clean decompression
## Summary

After `grove::deserialize()` successfully reads all compressed data, the underlying `std::ifstream` has its failbit set. This makes post-deserialization stream checks like `ifs.good()` or `!ifs.fail()` return failure even though decompression completed without errors.

## Reproduction

```cpp
std::ifstream ifs("index.ggx", std::ios::binary);
// ... read header + registries ...
auto grove = grove_type::deserialize(ifs);  // succeeds, no exception
assert(!ifs.fail());  // FAILS — failbit is set
```

## Expected Behavior

After `grove::deserialize()` returns successfully, the underlying stream should not have failbit set. It's acceptable for eofbit to be set (the grove is the last payload), but failbit implies a read error occurred.

## Context

The CLAUDE.md notes: "The `inflate_streambuf` seeks back unconsumed bytes on `Z_STREAM_END`, so multiple compressed payloads can be concatenated in one stream." This seek-back may be setting failbit when the compressed data ends exactly at EOF (no unconsumed bytes to seek back from).

## Current Workaround

Callers skip stream state checks after deserialization, relying on `grove::deserialize()` throwing on actual errors. This works but prevents callers from detecting stream corruption after the grove payload.

## Impact

- Stream state checks after `grove::deserialize()` are unreliable
- Discovered during serialization roundtrip tests in [ylab-hi/atroplex#17](https://github.com/ylab-hi/atroplex/pull/17)

riasc:
### Fixed
- `inflate_streambuf` leaves underlying stream with failbit set after successful decompression when the compressed data is the last payload in the file

Closes #301

## QC
- [x] Existing grove serialization tests pass (all 15 CI jobs green)
- [x] Underlying stream `!fail()` after `grove::deserialize()` when grove is last payload in file
- [x] Trailing data readable after grove deserialization (sentinel test)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed end-of-decompression stream handling so leftover error flags are cleared, ensuring subsequent reads/seeks work reliably.
* **Tests**
  * Added tests confirming deserialization leaves streams usable and trailing sentinel data can be read after roundtrips.
* **Changelog**
  * Documented the behavioral fix for stream state after decompression.


riasc:
## Code Review

**No bugs or guideline violations found.** The fix is correct and minimal — one line moved.

### Analysis

**Root cause**: When `inflate()` returns `Z_STREAM_END` with `avail_in == 0` (all compressed bytes consumed), the `source_.clear()` was inside the `if (avail_in > 0)` branch and never executed. The failbit from the last partial `source_.read()` persisted on the underlying stream.

**Fix correctness**: Moving `source_.clear()` before the conditional is safe in all cases:
- `avail_in > 0`: clear + seekback (same as before, just reordered)
- `avail_in == 0`: clear only (new — fixes the bug)
- Stream not at EOF (e.g., stringstream): `clear()` is a no-op when no error bits are set

**Test coverage**: Two new file-based tests in `tests/structure/serialization_test.cpp`:
1. `FileStreamStateCleanAfterDeserialize` — grove as last payload, asserts `!ifs.fail()`
2. `FileStreamStateCleanWithTrailingData` — grove + sentinel, asserts trailing reads work

Both use real `ifstream`/`ofstream` to exercise the partial-read codepath that `stringstream` doesn't trigger.

### Suggestion (optional)
Missing trailing newline at end of `serialization_test.cpp`. Minor, but some compilers warn about it.

Issue #306, opened 2026-04-11
Sorted insertion creates degenerate internal nodes (0 keys, 1 child)


riasc:
## Summary

`split_internal_node` uses `sorted_mid()` during sorted/bulk insertion, which produces degenerate right children with 0 keys and 1 child. These are structurally valid (children = keys + 1) but serve no routing purpose — they're just pointless wrappers.

## Root Cause

With order=3 and `sorted_mid = max(1, (order-1) * fill_factor) = 2`:

When an internal node overflows (3 keys, 4 children), `split_internal_node(mid=2)`:
- Left keeps keys[0..1] (2 keys) + children[0..2] (3 children)
- key[2] is promoted to parent
- Right gets keys[3..end] = **empty** + children[3] (1 child)

The right child is an internal node with 0 keys and 1 child.

This happens because `sorted_mid` packs everything into the left, which makes sense for **leaf** splits (append to rightmost) but creates degenerate nodes for **internal** splits.

## Proposed Fix

Use `default_mid()` (or a dedicated internal mid) for internal node splits instead of `sorted_mid()`. This would give a balanced 2-2 split for order=3 instead of 3-1.

Alternatively, `split_node()` could select the mid based on whether the child is a leaf or internal.

## Impact

- The degenerate nodes work correctly (the catch-all last child mechanism handles routing)
- But they waste memory and add unnecessary tree depth
- Found during #304 (B+ tree key removal) — the removal validator flagged these nodes

Discovered in #305.

riasc:
Fixed by #305. That PR removed `sorted_mid()` / `default_mid()` entirely and made both leaf and internal splits use a single `split_mid = order / 2` — the one midpoint that satisfies both leaf and internal minimum-occupancy constraints, so the degenerate `(0 keys, 1 child)` right child can no longer be produced. `fill_factor` was also removed as part of that change.

Not auto-closed because #305's `Closes` line omitted this issue.

Issue #304, opened 2026-04-11
Implement B+ tree key removal with rebalancing


riasc:
## Summary

Implement `remove_key()` on the B+ tree (grove) with proper rebalancing (node merging/redistribution on underflow).

## Motivation

Atroplex uses tombstone flags (`absorbed = true`) on segments during its ISM absorption phase. These tombstoned entries remain in the tree after build:

- Dead segments in leaf nodes
- Dead graph edges (SEGMENT_TO_EXON, EXON_TO_EXON) with the tombstoned segment's edge ID
- Wasted space in the serialized `.ggx` index

At 1000 samples: ~105K tombstoned segments out of ~15.5M total (~0.7%). The associated dead edges add up to ~840K out of 127M.

Currently there is no way to remove entries from the tree after insertion. A post-build compaction pass that removes tombstoned segments (and their associated graph edges) would clean up the index.

## Proposed API

```cpp
// Remove a key from the tree. Returns true if found and removed.
// Handles leaf underflow via redistribution or merging with siblings.
bool remove_key(const std::string& index_name, key<K, D>* key_to_remove);
```

Also needed: a way to remove edges associated with a key, or remove edges by predicate:

```cpp
// Remove all edges where the edge metadata matches a predicate
void remove_edges_if(auto predicate);

// Or: remove all edges originating from a specific key
void remove_edges_from(key<K, D>* source);
```

## Use case

After atroplex build completes:

```cpp
for each leaf in tree:
    for each segment key:
        if segment.absorbed:
            grove.remove_edges_from(key);  // clean up dead edges
            grove.remove_key(index_name, key);  // remove from tree + rebalance
```

This would produce a clean, compact tree for serialization.

## Complexity

Standard B+ tree deletion with rebalancing:
- Leaf deletion: O(log N) to find + O(B) to shift within leaf
- Underflow handling: redistribute from sibling or merge nodes
- Propagate merges up to internal nodes if needed
- Worst case: O(log N) node merges up to root

## Alternatives considered

- **Skip on serialize**: can't leave gaps in B+ tree structure without rebuilding
- **Rebuild tree on serialize**: works but wasteful — builds entire tree twice
- **Live with tombstones**: ~0.7% bloat is small but grows with more samples and creates dead edges that pollute edge traversal counts

riasc:
[...]

PR #308, opened 2026-04-13
Restore asymmetric sorted-leaf split

### LINKED ISSUE #285: split_internal_node(): sorted_mid() can produce a sibling with zero keys
## Summary

When `sorted_mid()` returns a value equal to `order - 1` (e.g., with a high `fill_factor`), `split_internal_node()` moves zero keys into the new right-sibling node. This produces an internal node whose `keys` vector is empty, which can cause an out-of-bounds read in `grove_query.ipp` during the next interval query that reaches that sibling.

Similarly, `default_mid()` returns `(order + 1) / 2`, which for the default `order == 3` is 2, and can leave the right sibling with zero keys in edge cases.

## Location

- `include/genogrove/structure/grove/grove_insert.ipp` — `default_mid()`, `sorted_mid()`, `split_internal_node()`
- `include/genogrove/structure/grove/grove_query.ipp` — Line 81 (reads past end if sibling has no keys)

## Suggested Fix

Clamp the internal split midpoint so it satisfies `1 <= mid <= order - 2`, guaranteeing at least one key on both sides after a split. A separate internal-split helper (distinct from the leaf-split midpoint) would make the invariant explicit and testable.

## Notes

- This is **pre-existing behavior** that predates PR #280 (the refactor PR that reorganized `grove.hpp` into `.ipp` files). It was not introduced by that refactor.
- Flagged during review of PR #280 (#280) comment: https://github.com/genogrove/genogrove/pull/280#discussion_r3003327577
- Requested by @riasc

riasc:
### Fixed
- **Sorted-insert hot-path regression from #305**: PR #305 replaced the sorted-insert leaf split with the symmetric `split_mid() = order/2`, citing "`fill_factor` had essentially no effect". That was wrong for the sorted-append hot path: with the old asymmetric split (`mid = order - 1`) the right sibling started with 1 key and absorbed `order - 2` more appends before the next split; with the symmetric split the right sibling starts at `order/2` and splits happen twice as often. The gh-pages benchmark chart showed **~2.07× regression at order=5** and 1.3–1.4× at order=10.
- **Reintroduce asymmetric split only on the sorted path**: new `split_leaf_node_sorted()` with `mid = order - 1`; `split_node()` gains a `bool sorted_append` parameter. Regular `insert()` / `insert_iter()` pass `false` and still use the symmetric split. **Internal splits always stay symmetric** regardless of `sorted_append`, so the degenerate `(0 keys, 1 child)` right internal from #306 cannot come back.
- **Relax min-occupancy for the rightmost leaf** in `validate_tree_invariants()` and the local validator in `grove_remove_test.cpp`. The sorted-append split leaves the rightmost leaf with as few as 1 key until the next `order - 1` appends fill it up; this is consistent with standard right-biased B+ trees where the rightmost leaf is the growth frontier and is exempt from min-occupancy. Removal rebalance (borrow / merge) still handles rightmost underflow correctly.

### Changed
- **`MergeLeaves` test** switched to regular (unsorted) `insert_data()`. It was previously written around the symmetric split's 2/2 layout at order=4 with 4 keys; with the asymmetric sorted split that's 3/1, and a single removal no longer underflows. The test's purpose is to exercise merge + root collapse, not sorted-insert semantics, so using regular insert here is the correct scope.

## Local verification

Debug benchmark via the `genogrove-gcc13-dev` container. Sorted / unsorted time ratios (lower is better; sorted should be much faster than unsorted):

| bench | ratio before #305 (Release, gh-pages) | ratio after #305 (Release, gh-pages) | ratio after this PR (Debug, local) |
|---|---|---|---|
| grove_creation/5000/5 | 0.46 | **0.96** | **0.56** |
| grove_creation/5000/10 | — | — | 0.57 |
| grove_creation/5000/15 | 0.45 | 0.63 | 0.50 |
| grove_creation/1000/5 | 0.43 | 1.17 | 0.64 |
| grove_creation/500/5 | 0.49 | 1.31 | 0.71 |

Sorted is firmly back to being roughly half the time of unsorted, which is the pre-regression profile. Debug vs Release accounts for most of the residual gap; the Release CI run on this PR will publish the actual recovered numbers to [the continuous benchmark chart](https://genogrove.github.io/genogrove/dev/bench/).

## Context

- The regression is internal-only performance — all of #305's correctness fixes (closes #285, #304, #306) are preserved.
- Internal splits stay symmetric, so #306 (degenerate `(0-key, 1-child)` right internal) cannot re-emerge.
- The rightmost-leaf exemption is a universal relaxation in the validator; it applies to trees in any state, not just those built via sorted insert.

Follow-up to #305.

## QC
- [x] All existing `grove_remove_test` tests pass (including `MergeLeaves`, `RootCollapse`, `BorrowFromLeftSibling`, `BorrowFromRightSibling`, stress tests)
- [x] All existing typed grove tests in `key_type_grove_test.hpp` pass with the relaxed rightmost-leaf min-occupancy
- [x] CI benchmark run publishes recovered Release numbers to gh-pages (sorted/unsorted ratio at order=5 should drop from ~0.96 back toward ~0.46)
- [x] \`split_leaf_node()\` (symmetric) is still called for regular inserts (grep: 2 call sites with \`sorted_append=false\`)
- [x] \`split_leaf_node_sorted()\` is called for sorted inserts (grep: 4 call sites with \`sorted_append=true\`)
- [x] Internal splits always use \`split_internal_node()\` (symmetric) — no new path can produce the #306 degenerate internal

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored sorted insertion performance by refining how leaf nodes split during sequential batch-append operations, which previously caused a measurable regression.

* **Tests**
  * Updated tree validation to properly account for the optimized leaf-splitting behavior.


riasc:
## Review summary

Traced through the correctness carefully — no bugs or guideline violations found. Notes:

- **`split_leaf_node_sorted` is only ever called at the catch-all position**: both `insert_sorted()` and the bulk sorted append route overflow into the rightmost spine, so \`child_index == parent.children.size() - 1\` always, and the \`parent.keys.insert(begin + index, parent_key_ptr)\` is an append to the end — never a shift that could narrow an existing separator.
- **Internal splits stay symmetric regardless of \`sorted_append\`**: #306's degenerate \`(0 keys, 1 child)\` right internal cannot come back. This is the load-bearing invariant for keeping this fix simple.
- **Removal of a 1-key rightmost leaf** rebalances correctly via borrow from the full left sibling — left has \`order - 1\` keys after an asymmetric split, which is always above \`leaf_min\` for \`order >= 3\`. Verified mentally for order=3, 4, 5.
- **The rightmost-leaf exemption** in the validator is a universal relaxation, not a \"sorted-insert-only\" exemption. That's correct: any tree in any state may have a \`< leaf_min\` rightmost leaf after the new sorted path, and the validator shouldn't care how the state was reached.

### Suggestion (optional)

- **Code duplication between `split_leaf_node` and `split_leaf_node_sorted`** (grove_insert.ipp:291-356). The two functions are identical except for the \`mid\` value. Could factor out a \`split_leaf_node_impl(parent, child, new_child, index, index_name, mid)\` and have the public variants just forward with their respective \`mid\`. ~15 lines saved. Not urgent — CLAUDE.md says to avoid abstractions beyond what the task requires, and the duplication is tiny. Flagging for future cleanup.

- **Validator duplication between `key_type_grove_test.hpp` and `grove_remove_test.cpp`**. The exemption logic is now in two places. Consolidating would be a larger refactor and is out of scope for this PR. Flagging as a future cleanup.

riasc:
Second review pass (covering d3992cf consolidation): all 14 compiler/platform builds pass against the new SHA, CodeRabbit pass, Performance Benchmarks pass (8m15s). Traced the consolidation — `split_leaf_node(..., int mid)` preserves behavior for both midpoints, and the shared `tree_validator.hpp` recursive walker matches the old per-file versions verbatim (minus the unused `leaf_count`). No issues.

Issue #86, opened 2026-01-22
Genogrove not thread-safe


riasc:
`insert_data` and other operations need to be made thread-safe for parallelization..

riasc:
## API design audit finding: graph_overlay static empty vector (2026-03-02)

`graph_overlay::get_edge_list()` at `include/genogrove/structure/grove/graph_overlay.hpp:162-163` uses a `static const std::vector<edge> empty_edges` as the fallback when no edges exist for a source key:

```cpp
const std::vector<edge>& get_edge_list(gdt::key<key_type, data_type>* source) const {
    static const std::vector<edge> empty_edges;
    auto it = adjacency.find(source);
    return (it != adjacency.end()) ? it->second : empty_edges;
}
```

While the vector is `const` (so no data race on the object itself), the static local initialization has thread-safety implications — C++11 guarantees thread-safe initialization of function-local statics, but this pattern is unusual and worth noting for any future thread-safety work.

Consider replacing with `std::span<const edge>` (empty span when no edges) which avoids the static entirely, or storing an empty member vector.

riasc:
## Design analysis for parallel on-the-fly construction (deferring implementation)

Target use case: parse GFF/BAM/BED once, build the grove in parallel per chromosome *without* pre-buffering the whole file.

### Proposed architecture: producer + per-chromosome consumers

```
[reader thread]                   ┌── worker chr1 ──→ grove.root_nodes["chr1"]
      │                           │
      ├─ entry → dispatcher ──────┼── worker chr2 ──→ grove.root_nodes["chr2"]
      │                           │
      ▼                           └── worker chrN ──→ grove.root_nodes["chrN"]
     EOF → close queues
```

- One reader thread drives `file_reader::read_next()` sequentially (no change to reader API).
- A dispatcher owns an MPSC queue per chromosome and routes each entry by `entry.chromosome`.
- One worker per chromosome drains its queue and calls `insert_data("chrN", ..., sorted)`. Because each worker is the only writer to its tree, **no per-tree locking is needed**.

### Grove-side state audit

The grove has five pieces of mutable state. All are currently shared across indices:

| State | Today | Needed for lockless parallel |
|---|---|---|
| `root_nodes` map | `unordered_map`, mutated per-index on first insert | Pre-populate before workers start |
| `rightmost_nodes` map | Same | Pre-populate; splits mutate the value, safe under exclusive worker |
[...]

PR #317, opened 2026-05-14
Make registry dedup-on-insert via intern() (#316)

### LINKED ISSUE #316: Add gdt::interning_registry<T> — dedup-on-insert sibling of gdt::registry<T>
## Motivation

`gdt::registry<T>` is a non-deduplicating singleton store — every `register_data(T)` call appends a new entry, even for duplicate values. That's the right shape for things like sample metadata (`gdt::registry<sample_info>` used in atroplex), where each entry is unique by construction.

But for **string interning** — where the same value is seen thousands of times and we want to collapse references to a single `uint32_t` ID — there's no upstream primitive. Consumers re-implement the same pattern over and over:

```cpp
class some_string_registry {
    std::unordered_map<std::string, uint32_t> str_to_id_;
    std::vector<std::string> id_to_str_;
    std::mutex mtx_;
public:
    uint32_t intern(const std::string& s) {
        std::lock_guard lock(mtx_);
        auto it = str_to_id_.find(s);
        if (it != str_to_id_.end()) return it->second;
        uint32_t id = id_to_str_.size();
        str_to_id_[s] = id;
        id_to_str_.push_back(s);
        return id;
    }
    const std::string& resolve(uint32_t id) const { return id_to_str_[id]; }
    // ... size / clear / serialize / etc.
};
```

[atroplex](https://github.com/ylab-hi/atroplex) currently maintains three such classes (`transcript_registry`, `gene_registry`, `source_registry` in `include/build/genomic_feature.hpp`) — all ~50 LOC of near-identical boilerplate. The forthcoming atroplex parallel-build work needs these to be thread-safe; pushing the primitive upstream means:

- One implementation, one test suite, one thread-safety story
- Future consumers (or other atroplex registries) reuse it instead of duplicating

## Proposed API

```cpp
namespace genogrove::data_type {

template<typename T>
class interning_registry {
public:
    using id_type = uint32_t;
    static constexpr id_type null_id = std::numeric_limits<id_type>::max();

    static interning_registry& instance();

    interning_registry(const interning_registry&) = delete;
    interning_registry& operator=(const interning_registry&) = delete;
    interning_registry(interning_registry&&) = delete;
    interning_registry& operator=(interning_registry&&) = delete;

    /// Return the ID for `value`, allocating one if first seen.
    /// Idempotent: intern(x) == intern(x) for all x.
    /// Thread-safe: serializes through an internal mutex.
    [[nodiscard]] id_type intern(const T& value);

    /// Reverse lookup. Throws std::out_of_range on invalid id.
    const T& get(id_type id) const;

    [[nodiscard]] bool contains(id_type id) const noexcept;

    [[nodiscard]] size_t size() const noexcept;
    [[nodiscard]] bool empty() const noexcept;

    void clear();
    static void reset();   // convenience for tests

    void serialize(std::ostream& os) const;
    [[nodiscard]] static interning_registry& deserialize(std::istream& is);

private:
    interning_registry() = default;
    std::unordered_map<T, id_type> lookup_;
    std::deque<T>                  storage_;   // matches gdt::registry's deque choice
    mutable std::mutex             mtx_;
};

}  // namespace genogrove::data_type
```

Mirrors `gdt::registry<T>`'s shape — same `id_type`, same `null_id` sentinel, same Meyer singleton, same `reset()` helper, same serialization trait requirement on `T`. The only structural addition is the lookup map and the mutex.

## Thread safety

Match the v0.22.0+ pattern. Acquire the internal mutex inside `intern()` and `clear()`. **Reads (`get`, `contains`, `size`, `empty`, `serialize`) stay unlocked** — `std::deque` guarantees pointer/reference stability across `push_back`, and the lookup map is monotonic (entries never removed except via `clear()` which is already locked). A release on lock unlock is sufficient for visibility; if TSAN flags it under strict mode, an alternative is `std::shared_mutex` with `shared_lock` on reads.

## Requirements on `T`

- Hashable (`std::hash<T>` specialization)
- Equality-comparable (`operator==`)
- Copyable or movable
- Whatever `serialization_traits<T>` requires (matching existing `gdt::registry<T>`)

`std::string` satisfies all; first use case in atroplex is three string-keyed registries.

## Relationship to existing `gdt::registry<T>`

Sibling, not replacement. Different semantics suit different consumers:

| Use case | Right tool |
|---|---|
| Per-entity metadata (sample_info, etc.) — each `register_data` should be a new entry | `gdt::registry<T>` |
| String/value pool — same value seen 1000× times, want one ID | `gdt::interning_registry<T>` |

The two could share a private base if it ends up cleaner; from the outside they remain separate types.

## Implementation sketch

`std::unordered_map<T, id_type>` for the lookup, `std::deque<T>` for stable-reference storage. Serialize as the storage vector (count + each entry); deserialize rebuilds the map by iterating. Wire format compatible with existing `gdt::registry<T>` if `T` matches, so `gdt::registry<std::string>` data could be loaded into `gdt::interning_registry<std::string>` if needed (though the reverse can't preserve dedup).

## Consumer rollout (atroplex side)

After this lands, atroplex's three registries become thin wrappers:

```cpp
// transcript_registry: pure alias
using transcript_registry = gdt::interning_registry<std::string>;

// gene_registry: dedup on gene_id, side-vector for metadata triplet
class gene_registry {
    gdt::interning_registry<std::string> ids_;
    std::vector<gene_info> metadata_;
public:
    uint32_t intern(const std::string& gid, const std::string& gname, const std::string& gbiotype) {
        size_t before = ids_.size();
        uint32_t idx = ids_.intern(gid);
        if (idx == before) metadata_.push_back({gid, gname, gbiotype});
        return idx;
    }
    const gene_info& resolve(uint32_t idx) const { return metadata_[idx]; }
};

// source_registry: cap at 16, narrow to uint8_t bit position, bitfield helpers
class source_registry {
    gdt::interning_registry<std::string> ids_;
    static constexpr size_t MAX_SOURCES = 16;
public:
    uint8_t intern(const std::string& s) {
        auto id = ids_.intern(s);
        if (id >= MAX_SOURCES) throw std::overflow_error{...};
        return static_cast<uint8_t>(id);
    }
    uint16_t mask(const std::string& s) { return 1u << intern(s); }
    // ... for_each(bitfield, fn), count(bitfield) ...
};
```

Net atroplex impact: ~-100 LOC of registry code, ~+40 LOC of wrappers, all thread-safety inherited from upstream.

riasc:
## Summary

Closes #316.

The issue proposed adding a sibling class `gdt::interning_registry<T>` for dedup-on-insert string interning, leaving `gdt::registry<T>` as the non-deduplicating primitive. After working through the trade-offs in conversation, we landed on a simpler design: **collapse to a single class**. The non-dedup `register_data` semantics had no real consumer in this repo, and a deduplicating registry strictly subsumes the use cases without a meaningful downside (the lookup-map memory cost is small relative to grove-side savings from collapsing references).

### What changed

- **`register_data()` → `intern()`** (breaking). `intern(value)` returns the existing ID for a value already in the registry, otherwise allocates a new one. Idempotent: `intern(x) == intern(x)`.
- **`find(value) → std::optional<id_type>`** added for probing without inserting.
- **`registry_value` concept** constrains `T` to hashable + equality-comparable.
- **Mutable `get(id)` overload removed** — mutating an interned value would desynchronize the internal value→id lookup.
- **Thread-safety**: `intern` / `find` / `clear` / `serialize` / `deserialize` protected by an internal `std::mutex`. `get(id)` / `contains(id)` / `size()` / `empty()` are unlocked fast paths; `get(id)` is safe under concurrent `intern()` iff the caller obtained `id` via a happens-before edge (the natural publication pattern). This lets the forthcoming atroplex parallel-build use one upstream primitive instead of three handwritten per-string-pool classes.
- **Serialization wire format unchanged** — the lookup map is rebuilt from the deque on `deserialize()`.

### Design decisions, briefly

- **Why one class, not two?** The non-dedup `register_data` path had two arguments for keeping it: (a) `T` not hashable, (b) caller needs distinct IDs for equal values. Neither applies to the metadata types this registry actually stores. The breaking change is contained — only consumer affected is the test suite.
- **Why no `shared_mutex`?** `intern()` and `find()` both touch the map; in the workload that matters (atroplex parallel build) `intern` is the hot path, so letting parallel `find`s run doesn't buy much. Plain `std::mutex` keeps the implementation small. Easy to switch later if profiling shows contention.
- **Why unlocked `get(id)`?** `std::deque::push_back` keeps existing pointers/references stable, and `get(id)` only ever reads an element whose id was returned by a prior `intern()`. The publication discipline that's already required to pass an id between threads also provides the happens-before edge that makes the read race-free. Documented in the docstring.

### Test plan

- [x] `genogrove_data_type_tests` passes locally (the user builds — please verify before merge)
- [x] New tests cover: idempotent intern, interleaved-dedup intern, dedup on complex values, find on existing/missing, find doesn't insert, lookup rebuilt on deserialize, concurrent intern (8 threads × 1000 calls × 32 distinct values → exactly 32 unique entries, all ids round-trip)
- [x] No remaining `register_data` references in source (`grep` confirms only historical CHANGELOG entries)


riasc:
## Code Review — /review-pr

CI: 14/14 build+test matrix entries pass; Performance Benchmarks pending (typical). No inline review comments to respond to; CodeRabbit's only post on this PR is a rate-limit notice with no actual review.

### Findings

**Suggestion — exception safety in `intern()` (theoretical, low priority)**

`include/genogrove/data_type/registry.hpp:118-121`:

```cpp
auto id = static_cast<id_type>(storage.size());
storage.push_back(value);              // 1
lookup.emplace(storage.back(), id);    // 2
```

If step 2 throws (allocation failure on the map node, or a throwing hash/eq), step 1 has already mutated `storage` and we leak an orphan entry — the deque now contains `value` at position `id` but the lookup doesn't know about it. The next `intern(value)` would push another copy. For `std::string`/`int` (`noexcept` hash) the only realistic trigger is `bad_alloc`, which puts you in a bigger mess anyway. Fixable with a `try { ... } catch (...) { storage.pop_back(); throw; }` around step 2. Flagging as optional because the trigger is narrow.

**Suggestion — memory cost (future optimization)**

Each interned value is stored twice: once in `storage` (deque), once in `lookup`'s map node key. For a million-entry registry of long strings, that's a real doubling of registry memory. Since `std::deque` keeps element addresses stable across `push_back`, the map could key on `const T*` (pointer into the deque) with a custom hash/eq that derefs the pointer. Halves the key storage. Not in scope here, but worth recording for #86 follow-up.

**Note — duplicate-tolerant `deserialize`**

If a wire-format payload happens to contain duplicates (e.g., from a pre-#317 `register_data` file that wrote the same value twice), the rebuild on `deserialize` puts all entries in the deque but only the **first** occurrence gets a lookup entry — subsequent duplicates become orphan entries in `storage`, reachable by `get(id)` but not via `intern()`. This is the documented "old files load fine" behavior; just calling out that the inconsistency is silent. Behavior is correct given the user's "old files can be re-created" stance.

**Note — test hash combiner is weak**

`tests/data_type/registry_test.cpp:55-64`: `h1 ^ (h2 << 1) ^ (h3 << 2)` is collision-prone. Fine for the test scope (no test depends on hash quality) but a real consumer specializing `std::hash<SampleInfo>` should use a stronger combiner.

### CLAUDE.md / memory compliance

- ✓ `[[nodiscard]]` correctly applied on `intern`, `find`, `deserialize`, `contains`, `size`, `empty`; correctly omitted on `get` (simple getter returning a stored reference).
- ✓ Bare member names (`storage`, `lookup`, `mtx`) — matches codebase convention after the underscore-strip commit.
- ✓ CHANGELOG entries reference both `#316` and `#317` in the required format.
- ✓ No CLAUDE.md edits in this PR.
- ✓ C++20 concepts (`registry_value`) used to constrain the template.
- ✓ Standard error types: `std::runtime_error` for capacity, `std::out_of_range` for invalid id.

riasc:
## Code Review (re-run) — /review-pr

CI: 14/14 build+test matrix entries pass on the new HEAD (`f8f0acd`); Performance Benchmarks pending (typical). No new inline review comments; no new issue-level activity since the last review.

### New commit since last review
`f8f0acd Roll back deque entry on lookup.emplace failure in intern()` — addresses the exception-safety suggestion from the previous review pass. The wrap is correct: `lock_guard` still spans the try/catch so the rollback runs under the lock, `storage.pop_back()` is no-throw for any reasonable `T` so the catch can rethrow cleanly, and the registry now provides the strong exception-safety guarantee for `intern()` (no observable mutation if it throws). Comment explains the why ("orphaned value would be invisible to future intern() calls") rather than the what, matching project style.

No new test for the failure path — the trigger requires either an injecting allocator or a throwing-hash type, which is more scaffolding than the narrow window warrants. Open to adding one if you want explicit coverage.

### Findings (this pass)

None. No bugs, no convention violations, no further suggestions.

### Project conventions

- ✓ `[[nodiscard]]` policy: present on `intern`, `find`, `deserialize`, `contains`, `size`, `empty`; correctly omitted on `get` (simple getter returning a stored reference).
- ✓ Bare member names (`storage`, `lookup`, `mtx`).
- ✓ CHANGELOG entries under `[Unreleased]` reference both `#316` and `#317` in the required format.
- ✓ C++20 concepts (`registry_value`) constrain the template.
- ✓ Standard error types: `std::runtime_error` for capacity, `std::out_of_range` for invalid id.
- ✓ Thread-safety contract documented in the class docstring and per-method notes.

### QC checklist

- ☐ `genogrove_data_type_tests` passes locally — left for the user (explicitly your call before merge).
- ✓ New tests cover the listed scenarios (verified earlier; still green on CI).
- ✓ No remaining `register_data` references in source.

### Label
`enhancement` — confirmed, matches the diff (changes existing functionality + adds `find`).

PR #320, opened 2026-05-15
Add `Tag` template parameter to `registry<T>` for distinct singletons

### LINKED ISSUE #319: Add tag template parameter to `registry<T>` for distinct singletons across same-T pools
## Problem

`gdt::registry<T>` is a singleton per type `T`. When two unrelated consumers in the same binary both pick the same `T` for different conceptual pools, they silently share the singleton:

```cpp
// Module X — transcript ids
auto id1 = gdt::registry<std::string>::instance().intern("ENST00000001"); // → 0
auto id2 = gdt::registry<std::string>::instance().intern("ENST00000002"); // → 1

// Module Y — biosample names, unaware of X
auto id3 = gdt::registry<std::string>::instance().intern("TCGA-A1-B2");   // → 2
auto id4 = gdt::registry<std::string>::instance().intern("ENST00000001"); // → 0 (!)

// Module X later:
const std::string& v = gdt::registry<std::string>::instance().get(2);
// expects: a transcript id from X's pool
// actual:  "TCGA-A1-B2" — Y's biosample name
```

No exception, no compile error, no UB — just wrong data. The bug is invisible until something downstream blows up. It manifests most easily with standard types (`std::string`, integers) where two independent components are likely to pick the same `T`, but it can happen with any shared type.

The current workaround is a strong-type wrapper per pool:

```cpp
struct transcript_id_value {
    std::string s;
    bool operator==(const transcript_id_value&) const = default;
};
template<> struct std::hash<transcript_id_value> {
    size_t operator()(const transcript_id_value& v) const noexcept {
        return std::hash<std::string>{}(v.s);
    }
};
using transcript_registry = gdt::registry<transcript_id_value>;
```

That works, but it forces ~6 lines of boilerplate per pool (struct + hash specialization), and call sites get noisier (`intern({\"ENST...\"})` instead of `intern(\"ENST...\")`, `.s` field access on `get()`). For atroplex specifically, this pattern is repeated for three different string-keyed pools — transcript ids, GFF source names, and (eventually) others — and the boilerplate adds up.

## Proposal

Add a phantom tag parameter to the template:

```cpp
template<registry_value T, typename Tag = void>
class registry { ... };
```

`Tag` is never instantiated — it's only used to make different template arguments produce different types, hence different singletons:

```cpp
using transcript_registry = gdt::registry<std::string, struct transcript_tag>;
using source_registry     = gdt::registry<std::string, struct source_tag>;
```

`registry<std::string, transcript_tag>` and `registry<std::string, source_tag>` are now distinct types with independent singletons. Call sites are unchanged from current ergonomics:

```cpp
transcript_registry::instance().intern("ENST00000001");  // → 0 in transcript pool
source_registry::instance().intern("HAVANA");            // → 0 in source pool (separate)
```

Type-safety bonus: passing a `transcript_registry::id_type` where a `source_registry::id_type` is expected becomes a compile error if the consumer aliases them separately, instead of a silent wrong-lookup.

## Backwards compatibility

`Tag = void` is the default, so every existing user of `gdt::registry<T>` becomes `gdt::registry<T, void>` at the type level — same template instantiation, same singleton, same behavior. No source changes required in any current consumer. No ABI break for header-only usage. Serialization format unchanged (the tag doesn't appear in the wire format).

## Implementation sketch

The change is one line in `include/genogrove/data_type/registry.hpp`:

```diff
-template<registry_value T>
+template<registry_value T, typename Tag = void>
 class registry {
   ...
 };
```

The `Tag` parameter appears nowhere in the body — it's a pure type-level discriminator. No mutex, storage, lookup, or serialization logic needs to change.

Documentation update in the class doc comment to mention the tag pattern, plus one test demonstrating that `registry<T, TagA>` and `registry<T, TagB>` have independent state.

## Alternatives considered

1. **Consumer-side strong-type wrappers** (status quo workaround). Works without upstream changes, but ~6 lines of boilerplate per pool, noisier call sites, and the pattern has to be rediscovered by every new consumer. Doesn't scale beyond a few pools.

2. **Runtime-keyed registries** (e.g., `instance(string_id)` returning per-id instances). Loses type safety, adds runtime overhead, complicates serialization. Rejected.

3. **Drop the singleton model entirely** and require callers to pass `registry&` around. Most flexible, but invasive refactor of every current consumer. Out of scope for this issue.

The tag-param approach is the smallest change that eliminates the footgun without requiring consumers to know about it.

## Motivation / downstream context

atroplex (https://github.com/ylab-hi/atroplex) is porting its three hand-rolled interning registries (`transcript_registry`, `gene_registry`, `source_registry`) onto `gdt::registry<T>` now that v0.23.0 makes the upstream primitive thread-safe and dedup-on-insert (#316 / #317). Two of those three pools are keyed on `std::string`, so the singleton-collision risk applies directly. With the tag parameter, atroplex's port becomes three one-line `using` aliases instead of ~30 lines of wrapper boilerplate.

riasc:
## Summary

- Adds a phantom `Tag = void` template parameter to `gdt::registry<T>` so two unrelated pools sharing the same value type (e.g. transcript ids vs sample names, both `std::string`) get independent singletons / ID spaces instead of silently colliding on a shared singleton.
- `Tag` is type-level only — it never appears in the class body. No storage, no runtime cost, no impact on the serialization wire format.
- Default `Tag = void` preserves the existing `registry<T>` singleton: every current call site stays binary-equivalent and old serialized files load unchanged.
- Class-level doc updated to describe the tag pattern and show the `using transcript_registry = registry<std::string, struct transcript_tag>;` idiom from the issue.

Closes #319.

## Test plan

- [x] New `RegistryTest.TaggedRegistriesAreIndependentSingletons` test exercises:
  - distinct singleton object identity for `registry<std::string>`, `registry<std::string, transcript_tag>`, `registry<std::string, sample_tag>`
  - the same string interns to id `0` in two different tagged pools (no collision)
  - `clear()` on one tagged pool leaves the other two pools untouched
  - the default-tagged singleton stays empty when only tagged variants are used
- [x] All existing `RegistryTest.*` tests continue to pass (default-tag behavior is unchanged)
- [x] CHANGELOG `[Unreleased] > Added` entry references issue #319

## Summary by CodeRabbit

* **New Features**
  * Registry now supports an optional `Tag` template parameter (defaults to `void`) enabling independent singleton registries with separate ID spaces for identical value types while maintaining backward compatibility.

* **Documentation**
  * Updated registry documentation with examples of tagged registry usage.

* **Tests**
  * Added test case verifying independent singleton behavior across tagged registries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/genogrove/genogrove/pull/320?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- **`gdt::registry<T, Tag>`**: optional phantom `Tag` template parameter discriminates singletons that share the same value type `T`, so two unrelated pools (e.g. transcript ids vs sample names, both `std::string`) get independent ID spaces without consumers having to wrap `T` in a strong-type struct. `Tag` defaults to `void`, so every existing `registry<T>` keeps its singleton and call sites are unchanged. Tag is type-level only — no storage, no runtime cost, no impact on the serialization wire format. ([#319](https://github.com/genogrove/genogrove/issues/319))
+- **`gdt::registry<T, Tag>`**: optional phantom `Tag` template parameter discriminates singletons that share the same value type `T`, so two unrelated pools (e.g. transcript ids vs sample names, both `std::string`) get independent ID spaces without consumers having to wrap `T` in a strong-type struct. `Tag` defaults to `void`, so every existing `registry<T>` keeps its singleton and call sites are unchanged. Tag is type-level only — no storage, no runtime cost, no impact on the serialization wire format. ([#319](https://github.com/genogrove/genogrove/issues/319), [#320](https://github.com/genogrove/genogrove/pull/320))
 
 ## [0.23.0] - 2026-05-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **`gdt::registry<T, Tag>`**: optional phantom `Tag` template parameter discriminates singletons that share the same value type `T`, so two unrelated pools (e.g. transcript ids vs sample names, both `std::string`) get independent ID spaces without consumers having to wrap `T` in a strong-type struct. `Tag` defaults to `void`, so every existing `registry<T>` keeps its singleton and call sites are unchanged. Tag is type-level only — no storage, no runtime cost, no impact on the serialization wire format. ([#319](https://github.com/genogrove/genogrove/issues/319))
+
 ## [0.23.0] - 2026-05-15
 
 ### Added
--- a/include/genogrove/data_type/registry.hpp
+++ b/include/genogrove/data_type/registry.hpp
@@ -39,7 +39,12 @@ concept registry_value =
 /**
  * @brief Singleton registry that interns values into small integer IDs.
  *
- * @tparam T The value type to intern. Must be hashable and equality-comparable.
+ * @tparam T   The value type to intern. Must be hashable and equality-comparable.
+ * @tparam Tag Phantom type used only to discriminate singletons. Different `Tag`
+ *             arguments produce distinct types with independent ID pools; the
+ *             default `void` preserves the original "one singleton per T"
+ *             behavior. `Tag` never appears in the body — no storage, no
+ *             serialization, no runtime cost.
  *
  * @details Every distinct value gets one stable ID; calling intern() with the same
  * value always returns the same ID. The point is to collapse many references to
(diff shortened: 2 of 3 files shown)


riasc:
## Code Review

### Conventions

- ✓ C++20 concept-constrained template parameter (`registry_value T`) kept intact; new `Tag` is an unconstrained `typename` since it's purely a type-level discriminator.
- ✓ Public class doc updated (`@tparam Tag`, expanded `@details`, additional `@code` example).
- ✓ Test added for the new behavior — covers singleton identity, same-string-different-pool id allocation, isolated `clear()`, and untouched default-tag singleton.
- ✓ Cross-pool `EXPECT_NE` between `registry<std::string>` and `registry<std::string, Tag>` uses `static_cast<void*>` (the two pointer types are unrelated, so the cast is required to compare).
- ✓ Test resets the tagged singletons at both start and end (the fixture's `SetUp`/`TearDown` only resets the default-tag `registry<std::string>`).
- ✓ CHANGELOG `[Unreleased] > Added` entry references both issue and PR.
- ✓ Naming follows project convention (no Google-style trailing-underscore members, bare snake_case).
- ✓ No coupling between `registry` and any grove-side type — `Tag` lives entirely in `registry`.

### Bugs

None.

### Guideline violations

- ~~CHANGELOG entry only referenced the issue, not the PR.~~ Fixed in 37ad948 — both `#319` and `#320` are now linked per the established `([#issue], [#pr])` format.

### Suggestions

None worth blocking on. The diff matches the implementation sketch in #319 almost line-for-line; the only judgment call (using inline `struct transcript_tag` / `struct sample_tag` in the test) is the idiomatic phantom-tag pattern.

### Existing comments

- **CodeRabbit (auto-summary)**: no actionable comments — *Skip*. The CodeRabbit walkthrough is informational only; its docstring-coverage warning is irrelevant here (a one-line template signature change with a documented `@tparam` does not need a separate docstring).

### CI

All 15 build matrix entries green (gcc 13/14, clang 16/17/18 ×Debug/Release on Ubuntu 24.04; apple-clang 15/16 on macOS 14/15) plus performance benchmarks.

### Label

`feature` — adds a new (opt-in) template parameter and capability.

PR #391, opened 2026-05-16
bam_reader: stop fabricating intervals for zero-ref CIGARs

### LINKED ISSUE #327: [High] bam_reader::compute_interval fabricates 1-base interval for zero-ref-consuming CIGARs
## Finding

`src/io/bam_reader.cpp:273-276` — Pure soft-clip ("100S") and hard-clip-only secondary records consume zero reference bases; `compute_interval` silently coerces `ref_len = 0 -> 1`, fabricating a `[pos, pos+1)` interval that doesn't reflect any real reference span.

## Risk

Produces spurious overlap hits during `intersect()`.

## Fix

Pick a contract and document it: either (a) propagate `start == end` truthfully, (b) drop these reads (similar to `BAM_FUNMAP`), or (c) explicit caller-side filter. Update doxygen on `sam_entry.start`/`end`. Add a regression test pinning the chosen semantics.

## Source

riasc:
### Fixed
- **`bam_reader::compute_interval` no longer fabricates a 1-base interval for zero-ref-consuming CIGARs** (**behavior change**): pure soft-clip (`100S`) and hard-clip-only secondary records now propagate `start == end == POS` truthfully instead of being silently coerced to `[pos, pos+1)`. The old behavior produced spurious overlap hits during `intersect()` queries. Adds `sam_entry::consumes_reference()` as the clean caller-side gate before converting to a closed `gdt::interval(start, end - 1)` — the conversion is only valid when `start < end`. Mirrors the existing pattern where unmapped reads pass through with `start == end == 0`. ([#327](https://github.com/genogrove/genogrove/issues/327))

Closes #327

## Migration

Code that builds a `gdt::interval` from `sam_entry::start`/`end` should now check `consumes_reference()` (or `start < end`) before inserting into a grove:

```cpp
for (const auto& entry : reader) {
    if (!entry.consumes_reference()) continue;   // skip pure-soft-clip / hard-clip-only
    grove.insert_data(entry.chrom, gdt::interval(entry.start, entry.end - 1), entry);
}
```

Code that consumed the previous fabricated 1-base interval will now silently drop those records when filtered as above — which is the intended behavior (those records were never legitimate ref-overlap hits).

## QC
- [x] New `BamReaderTest` cases pass: `PureSoftClipDoesNotFabricateInterval`, `ConsumesReferenceTrueForNormalRecord`, `HardClipOnlySecondaryDoesNotFabricateInterval`, `ConsumesReferenceFalseForUnmappedRead`
- [x] Existing `BamReaderTest` suite still passes (the change only affects zero-ref-consuming records; normal CIGARs are unchanged)

## Summary by CodeRabbit

* **New Features**
  * Added method to check if a BAM record consumes reference bases.

* **Bug Fixes**
  * Fixed BAM reader handling of CIGAR records consuming zero reference bases, preventing spurious overlap hits during intersection operations. Zero-consuming reads now produce zero-length intervals instead of fabricated intervals.

* **Tests**
  * Added test coverage for zero-reference-consuming CIGAR record handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/genogrove/genogrove/pull/391?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


riasc:
## Code Review

Reviewed the diff and CodeRabbit's status. Tight, scoped fix that matches the chosen contract.

### Bug

None.

### Guideline violation

None. `[[nodiscard]]` on the new helper is consistent with the existing `is_mapped()` / `is_primary()` / `cigar_string_repr()` on `sam_entry`. Closed-interval conversion guidance in the doxygen matches the project's I/O entry coordinate convention.

### Suggestion (optional)

- **Hard-clip-only test for symmetry.** The fix covers both pure soft-clip (`100S`) and hard-clip-only secondaries (e.g. `100H` on a `BAM_FSECONDARY` record) — they take the same `ref_len == 0` code path. The regression test only exercises `100S`. Adding one more test with a hard-clip-only secondary would pin both halves of the contract the issue describes. Not blocking; the underlying code path is identical.
- **End-to-end intersect() spurious-hit test.** Could insert a normal record at `chr1:1000-1050` plus a pure-soft-clip record at `chr1:999` into a grove, query `[995, 999]`, and assert the soft-clip record is filtered out by `consumes_reference()`. Belongs more in `grove_remove_test.cpp` style integration tests; out of scope for a reader fix.

### Comments processed

- **CodeRabbit issue-level**: rate-limit warning, no review performed. No actionable feedback to skip or address.
- No inline review comments.

### CHANGELOG

Added in `9fbb7cb` under `### Fixed` with both `[#327]` and `[#391]` links and the **(behavior change)** marker consistent with the previous `bam_reader::read_next()` entry.

### CI

16/16 jobs green on `d9a4f65` (full matrix + benchmarks).

riasc:
## Re-review — follow-up commit \`4480bcd\`

Only change since the prior review is the addition of \`HardClipOnlySecondaryDoesNotFabricateInterval\` in \`tests/io/bamfile-test.cpp\` — no production code touched.

### Findings
None. New test is well-formed:
- FLAG=256 (\`BAM_FSECONDARY\`) and CIGAR=\`100H\` exercise the same \`ref_len == 0\` code path as the soft-clip case but via a different op.
- Asserts \`start == end == 499\` (POS=500 → 0-based 499), \`consumes_reference() == false\`, plus \`is_mapped() == true\` / \`is_primary() == false\` to confirm the secondary-flag handling.
- Uses \`SEQ/QUAL=*\` correctly (hard-clipped portions are not present in SEQ per SAM spec).

### CI

All 16 matrix jobs pass on \`4480bcd\` (including the new test).

### CHANGELOG

Unchanged from \`9fbb7cb\` — same \`### Fixed\` entry covers this contract since the production fix is identical; the new test just pins the other half of the issue's described scenarios.

PR is ready to merge.

PR #392, opened 2026-05-16
bgzf_utils: preserve embedded NULs in next_data_line

### LINKED ISSUE #346: [Medium] bgzf_utils next_data_line: std::string(str.s) truncates at embedded NUL
## Finding

`src/io/bgzf_utils.cpp:23` — `std::string line(str.s)` constructs from C-string, truncating at any embedded `\0`.

## Risk

BGZF/gzip lines are normally text-only, but a malformed input could silently truncate.

## Fix

Use `std::string line(str.s, str.l);` (kstring tracks length in `.l`). The trailing `\r` strip already works with byte-counted strings.

## Source

riasc:
### Fixed
- **`bgzf_next_data_line` no longer truncates lines at the first embedded NUL** (**minor behavior change**): line construction switched from `std::string(str.s)` to `std::string(str.s, str.l)`. kstring is length-counted, not NUL-terminated, so the C-string ctor stopped at the first `\0` and silently dropped everything past it — including trailing tab-separated fields that downstream BED/GFF parsers needed. The trailing `\r` strip and empty/comment checks already operate on byte-counted strings, so the rest of the function is unchanged. Affects both `bed_reader` and `gff_reader` (both call this helper). ([#346](https://github.com/genogrove/genogrove/issues/346))

Closes #346

## QC
- [x] New `bedfileTest.embeddedNulInFieldPreserved` passes (synthetic BED6 with NUL in the name column round-trips with all 5 name bytes and the trailing score / strand fields)
- [x] Existing `bedfileTest` and `gfffileTest` suites still pass (the fix only changes behavior for lines with embedded NULs; NUL-free lines hit the same code path with identical output)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue in BED and GFF file parsing where fields containing embedded NUL bytes were incorrectly truncated. The parser now properly preserves the full content of fields containing NUL characters.

* **Tests**
  * Added regression test to verify embedded NUL bytes in file fields are properly preserved.

[Review Change Stack](https://app.coderabbit.ai/change-stack/genogrove/genogrove/pull/392?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


riasc:
## Code Review

Tightly scoped one-character fix (\`(str.s)\` → \`(str.s, str.l)\`) with a synthetic-input regression test that exercises both halves of the contract: the embedded NUL survives into the parsed name field AND the trailing tab-separated columns still parse.

### Bug

None.

### Guideline violation

None. \`std::ofstream(path, std::ios::binary)\` + explicit \`put('\\0')\` is the right way to embed a NUL on all platforms (text-mode \`\\0\` would otherwise be subject to translation on Windows; works on Linux/macOS regardless).

### Suggestion (optional)

Edge cases not in the test that all hit the same code path and would pass for free:
- NUL at the very end of a line (just before \`\\n\`).
- Multiple NULs in one line.
- NUL in a chromosome column (would propagate to \`entry.chrom\`).
- Gzipped BED with NUL.

None are worth adding to this PR — the existing test pins the bug. Listing them in case you ever want to extend coverage.

### Comments processed

- **CodeRabbit issue-level**: hit a rate limit and skipped its review (third PR in a row affected). No actionable feedback.
- No inline review comments.

### CHANGELOG

Added in \`e21c4c2\` under \`### Fixed\` with both \`[#346]\` and \`[#392]\` links.

### Branch state

Rebased onto current main (now includes the #393 CI trim) and force-pushed. The new CI run on \`e21c4c2\` will be the **first real-world validation of the trimmed pipeline** — expect 7 Debug jobs, ~1.5 min wall, no benchmarks/no Release.

### QC

Items 2, 3, 4 marked based on the pre-rebase full-matrix run (all 14 builds green). Since the only change since then is the CHANGELOG line, the trimmed Debug subset will pass too. Item 1 (human check) left for you.

PR #393, opened 2026-05-16
Trim PR CI: drop benchmarks and Release builds for faster feedback

riasc:
### Changed
- **Drop benchmarks from PR runs** (`continuous-benchmarking.yml`): the `pull_request` trigger is removed. PR-time benchmark runs were never pushed for baseline comparison, so the ~8 min job per PR was pure waste. The `push: main` trigger is kept, so the baseline still updates on each merge.
- **PR runs Debug only; main runs Debug + Release** (`ci-ubuntu.yml`, `ci-macos.yml`): the `build_type` matrix is now conditional on `github.event_name`. Halves the build matrix on PRs (14 → 7 jobs). Release-only failures are rare here (no LTO, `NDEBUG` only deletes asserts) and recent failures we've actually hit (#388, #391) were **Debug-only** — so we keep the bug-catching configuration on PRs and drop the one that almost never fails. Post-merge main runs still validate both build types.

Net effect per PR: 14 build jobs + 1 benchmark → 7 build jobs, no benchmark. Wall time ~10 min → ~1.5 min.

## QC
- [x] On this PR: only 7 jobs run (5 Ubuntu Debug + 2 macOS Debug), no `Performance Benchmarks` job, no `*-Release` jobs
- [x] All 7 PR jobs pass
- [x] Once merged: a manual push to main triggers the full matrix (14 build jobs + 1 benchmark)

### DIFF
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -12,7 +12,10 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-14, macos-15]
-        build_type: [Debug, Release]
+        # PRs run Debug only for fast feedback; main runs both. Release-only
+        # bugs are rare (no LTO here, NDEBUG only deletes asserts) and get
+        # caught post-merge on the full main-branch matrix.
+        build_type: ${{ github.event_name == 'pull_request' && fromJSON('["Debug"]') || fromJSON('["Debug", "Release"]') }}
         include:
           - os: macos-14
             clang_version: "15"
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -11,7 +11,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        build_type: [Debug, Release]
+        # PRs run Debug only for fast feedback; main runs both. Release-only
+        # bugs are rare (no LTO here, NDEBUG only deletes asserts) and get
+        # caught post-merge on the full main-branch matrix.
+        build_type: ${{ github.event_name == 'pull_request' && fromJSON('["Debug"]') || fromJSON('["Debug", "Release"]') }}
         c_compiler: [gcc-13, gcc-14, clang-16, clang-17, clang-18]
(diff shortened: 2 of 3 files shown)


riasc:
## Code Review

This PR is self-validating — the CI list itself proves the cuts work. Exactly 7 jobs ran (5 Ubuntu Debug + 2 macOS Debug), all green in ~1.5 min wall, no \`Performance Benchmarks\` job, no \`*-Release\` jobs.

### Bug

None.

### Guideline violation

None.

### Suggestion (optional)

The conditional matrix expression \`\${{ github.event_name == 'pull_request' && fromJSON('["Debug"]') || fromJSON('["Debug", "Release"]') }}\` uses GitHub Actions' ternary-equivalent pattern. Has a known footgun when the "if true" branch evaluates to a falsy value (empty string, 0, empty array) — the expression would short-circuit to the "else" branch instead. Doesn't apply here (\`fromJSON('["Debug"]')\` is a non-empty array, always truthy), so this is safe as written. Worth keeping in mind for future matrix expressions; an explicit \`if/else\`-style with separate \`include\`/\`exclude\` blocks avoids the pitfall but is more verbose. Not changing.

### Comments processed

- **CodeRabbit issue-level**: hit a rate limit and skipped the review. No actionable feedback. (Affects two PRs back-to-back now — worth noting.)
- No inline review comments.

### CHANGELOG

Skipped intentionally. \`ci\` label has no entry in the project's label → CHANGELOG section mapping, and Keep a Changelog convention is "for humans, not machines" — focused on consumer-visible behavior. Past CHANGELOGs have no CI-only entries. If you want this recorded for posterity, it would go under \`### Changed\` with a "Developer experience" or "CI" prefix, but I'd skip.

### QC

- Items 2 ("only 7 jobs run") and 3 ("all 7 pass") marked — verified by the live CI list.
- Item 4 ("once merged: push to main triggers full matrix") left **unchecked** — it's a post-merge validation step that can't be confirmed until after the squash-merge. You'll see it in the main-branch CI run after merging.

PR #394, opened 2026-05-17
file_reader_iterator: fix operator== for non-end iterators

### LINKED ISSUE #325: [High] file_reader_iterator::operator== returns true for any two non-end iterators sharing the same reader
## Finding

`include/genogrove/io/file_reader.hpp:83-90` — Two iterators that have advanced different numbers of times compare equal.

## Risk

STL algorithms relying on iterator-by-iterator equality (e.g. `std::distance` with input-iterator semantics) get wrong results. Hidden today only because users typically compare against `end()`.

## Fix

Document that only `it == end()` is meaningful (single-pass input iterator) and assert / forbid non-end equality, or compare `at_end_` flags + underlying `current_entry_` identity (e.g. via a monotonic record index).

## Source

riasc:
### Fixed
- **`file_reader_iterator::operator==` now respects iterator position** (**minor behavior change for an edge case**): two non-end iterators on the same reader used to compare equal unconditionally — the operator only checked `at_end_` + `reader_` identity. After this PR, two non-end iterators compare equal iff they have the same parent reader AND have advanced the same number of times (tracked via a new monotonic `pos_` counter, bumped on each successful `advance()`). Restores the input-iterator contract that "two iterators compare equal iff they refer to the same position." Without the fix, copying an iterator and advancing the original would leave the two comparing equal even though their `current_entry_` had diverged. Affects every reader (`bed_reader`, `gff_reader`, `bam_reader`, `fasta_reader`) since they all share the same `file_reader_iterator`. ([#325](https://github.com/genogrove/genogrove/issues/325))

Closes #325

## Impact

- End-iterator comparisons (`it == reader.end()`) are unchanged — same code path, same semantics. This is what 100% of range-for loops and STL find-style algorithms hit.
- The bug only manifested when the user explicitly copied an iterator and advanced the original. In normal idiomatic use (range-for + single iterator) it was invisible.
- Memory cost: 8 bytes per iterator. Iterators are stack-local and short-lived.

## QC
- [x] New `bedfileTest` iterator-equality tests pass: `iteratorEqualToSelf`, `iteratorCopyEqualToOriginalBeforeAdvance`, `iteratorCopyUnequalAfterOriginalAdvances`, `iteratorBeginUnequalToEnd`, `iteratorEndEqualsEnd`, `iteratorDrainedEqualsEnd`
- [x] Existing `bedfileTest`, `gfffileTest`, `bamfileTest`, `fastaReaderTest` suites still pass (the change only affects equality between two non-end iterators on the same reader; range-for loops and end-comparisons hit the same code path with identical output)


riasc:
## Code Review

Clean self-contained fix. The pre-existing \`iteratorPostIncrement\` test already exercised the copy-then-advance pattern but didn't assert equality between the two iterators — that's exactly the gap this PR fills.

### Bug

None.

### Guideline violation (self-caught, fixed in \`6f69283\`)

My initial test \`iteratorBeginUnequalToEnd\` called \`reader.begin()\` twice, which violates the project's documented single-pass convention (\`feedback_iterator_single_pass.md\` — "don't call reader.begin() twice; it consumes a record"). The test happened to pass because begin-vs-end equality is invariant regardless of which record was consumed, but the pattern was wrong. Restructured to a single \`begin()\` call.

### Suggestion (optional)

Test coverage is on \`bedfileTest\` since BED is simplest, and the iterator template is shared across all readers — so testing once is enough for confidence. Could add symmetric tests in \`gfffileTest\` / \`bamfileTest\` / \`fastaReaderTest\` for redundancy but it would just be copy-paste. **Skip** for this PR.

### Comments processed

- **CodeRabbit issue-level**: rate-limited again (fourth PR in a row). No actionable feedback. Worth re-evaluating CodeRabbit usage if the rate-limit pattern keeps blocking it.
- No inline review comments.

### CHANGELOG

Added in \`6f69283\` under \`### Fixed\` with both \`[#325]\` and \`[#394]\` links.

### CI

7/7 Debug jobs green on \`8a5edf8\` (the pre-fixup commit). CI is re-running now on \`6f69283\` with the trimmed pipeline — will mark items based on that result.

PR #395, opened 2026-05-17
inflate_streambuf: throw on seek failure for non-seekable sources

### LINKED ISSUE #323: [High] inflate_streambuf silently requires seekable source
## Finding

`include/genogrove/structure/grove/zlib_streambuf.hpp:142-164` — `underflow()` calls `source_.seekg(-avail_in, cur)` whenever the deflate stream ends with unconsumed buffered bytes.

## Risk

Requires source seekability; fails silently on pipes / non-seekable streambufs. The "concatenable payloads" claim in CLAUDE.md silently degrades on non-seekable inputs — trailing data is lost without an error.

## Fix

After `source_.seekg(...)` check `source_.fail()` and throw `std::runtime_error("inflate_streambuf: source stream is not seekable; concatenated payloads require a seekable source")` if it failed. Alternatively, expose an `unconsumed_bytes()` accessor so non-seekable callers can rewind manually.

## Source

Spotted during `/cpp-audit` on 2026-05-16. Related: #301

riasc:
### Fixed
- **`inflate_streambuf` no longer silently drops trailing bytes on non-seekable sources** (**minor behavior change**): after `Z_STREAM_END` with leftover buffered bytes, the streambuf rewinds them via `source_.seekg(-avail_in, cur)` so subsequent reads see the trailing data — that's how concatenated compressed payloads work. The seek result wasn't checked, so non-seekable sources (pipes, sockets, custom non-seekable streambufs) silently failed the rewind and dropped the trailer without any error, violating the "concatenated payloads" contract in CLAUDE.md. Now checks `source_.fail()` after the seek and throws `std::runtime_error("inflate_streambuf: source stream is not seekable; concatenated payloads require a seekable source")`. Only fires when there are leftover bytes to rewind (`avail_in > 0`) — existing single-payload usage on seekable streams (`std::ifstream`, `std::stringstream`) is unaffected. ([#323](https://github.com/genogrove/genogrove/issues/323))

Closes #323

## Impact

- **No effect** on normal file-based deserialization — `std::ifstream` is seekable, all existing tests pass.
- **No effect** on stringstream-based usage in tests — `std::stringstream` is seekable.
- **Surfaces previously silent data loss** for any code piping a `.ggx` file through `std::cin` or wrapping in a custom non-seekable streambuf. Such code now fails loudly instead of silently dropping trailing bytes.

## QC
- [x] New `SerializationTest.NonSeekableSourceWithTrailingDataThrows` passes — synthetic non-seekable streambuf wraps a serialized grove + 4-byte trailer; `deserialize()` throws `std::runtime_error`
- [x] Existing `SerializationTest` suite still passes — seekable-stream paths (`FileStreamStateCleanAfterDeserialize`, `FileStreamStateCleanWithTrailingData`) are unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Deserialization now properly detects and reports errors when non-seekable input sources (such as pipes or sockets) contain unconsumed trailing bytes after payload processing, preventing silent data loss in concatenated payload scenarios. This validation is conditional on the presence of leftover bytes. Single-payload usage on seekable streams remains unaffected by this change.

[Review Change Stack](https://app.coderabbit.ai/change-stack/genogrove/genogrove/pull/395?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


riasc:
## Code Review

Clean defensive fix. The check is narrowly scoped to the only path that can leak data silently (Z_STREAM_END with avail_in > 0), so seekable single-payload usage is untouched.

### Bug

None.

### Guideline violation (self-caught, fixed in \`613bd9d\`)

My initial test used \`EXPECT_THROW({ ... grove<gdt::interval, int>::deserialize(is); }, std::runtime_error)\`. The comma inside the template arg made the preprocessor split the macro into three args, breaking the build on every Debug job. Fixed by hoisting a \`using grove_t = ...\` alias above the macro. Same family of gotcha as the existing memory note on K-mer brace-init inside \`EXPECT_THROW\`.

### Suggestion (optional, skipped)

The issue listed an optional alternative — expose \`unconsumed_bytes()\` so non-seekable callers could rewind manually. Skipped: no caller needs it today, and adding unused public API surface is over-engineering. Can be added later if a real consumer (e.g. atroplex piping a \`.ggx\` over stdin) materializes.

### Comments processed

- **CodeRabbit issue-level**: rate-limited (fifth PR in a row affected — at this point I'd consider an issue against the bot config rather than treating it as per-PR noise).
- No inline review comments.

### CHANGELOG

Added in \`17af1cc\` under \`### Fixed\` with both \`[#323]\` and \`[#395]\` links.

### CI

7/7 Debug jobs green on \`613bd9d\` after the macro fix. CI is re-running now on \`17af1cc\` (CHANGELOG-only commit) — expecting the same.

Issue #332, opened 2026-05-16
[High] key::operator== requires-clause doesn't constrain data_type equality


riasc:
## Finding

`include/genogrove/data_type/key.hpp:299-306` — `operator==` requires only `std::equality_comparable<key_type>` but in the non-void branch also calls `data == other.data`.

## Risk

If `data_type` is not equality-comparable, fails deep in template instantiation rather than at the use site with a clean concept error.

## Fix

Tighten the clause: `requires std::equality_comparable<key_type> && (std::is_void_v<data_type> || std::equality_comparable<data_type>)`.

## Source

Spotted during `/cpp-audit` on 2026-05-16.

PR #397, opened 2026-05-17
query_result + flanking_query_result: const pointers + key_type_base concept

### LINKED ISSUE #324: [High] query_result / flanking_query_result expose mutable pointers + unconstrained template params
## Finding

`include/genogrove/data_type/query_result.hpp:52,87,101,109`; `include/genogrove/data_type/flanking_query_result.hpp:41,54,63,74,83` — Result containers store/return non-const `key<...>*` pointers, and both class templates use unconstrained `template<typename key_type, ...>`.

## Risk

Callers can do `result.get_keys()[0]->set_value(...)` and silently corrupt B+ tree ordering invariants. Unconstrained template params also mean misuse fails deep inside `key<>` instantiation rather than at the use site.

## Fix

(a) Change element type to `const key<key_type, data_type>*`. (b) Replace `template<typename key_type, ...>` with `template<key_type_base key_type, ...>` and `#include <genogrove/data_type/key_type_base.hpp>`.

## Source

Spotted during `/cpp-audit` on 2026-05-16. Related: #157, #113, #174

riasc:
### Changed
- **`query_result` / `flanking_query_result` store and return `const key<...>*`** (**breaking source change for any caller that bound result pointers to non-const `key*`**): pre-fix callers could do `result.get_keys()[0]->set_value(...)` and silently corrupt B+ tree ordering invariants. After this PR the const propagates through to callers at the type level so the mutation attempt is a compile error. Callers that legitimately need a mutating `key*` (rare; the only realistic use was the pointer returned by `insert_data()` itself) should hold on to that original pointer instead of re-deriving from results.
- **Template params constrained with `key_type_base` concept**: `template<typename key_type, ...>` → `template<key_type_base key_type, ...>` on both result containers. Misuse with a non-conforming type now fails at the container instantiation site with a clean concept error instead of deep inside `key<>` template instantiation. ([#324](https://github.com/genogrove/genogrove/issues/324))

Closes #324

## Impact

- **No effect** on internal grove code (`grove_query.ipp`, `grove_flanking.ipp`) — they construct results from non-const `key*` and implicit conversion to `const key*` happens at the API boundary.
- **No effect** on `cli/src/handlers/{bed,gff}.cpp` — they use `for (auto* k : result.get_keys())` where `auto*` now deduces to `const key*`; loop bodies only read via const methods.
- **No effect** on existing tests — none of them bound result pointers to non-const `key*` or called non-const methods on them.
- **Behavior change** for external callers who wrote `key* k = result.get_keys()[0];` — they get a compile error and must either widen to `const key*` (the common case) or `const_cast` (rare; deliberately loud).

## Test

Compile-time `static_assert` regression tests added to `query_result-test.cpp` pin the exact return types of `query_result::get_keys()`, `flanking_query_result::get_predecessor()`, and `flanking_query_result::get_successor()`. They fail at build time if the const drifts off.

## QC
- [x] New `query_result_test.get_keys_returns_const_pointer_vector` and `query_result_test.flanking_returns_const_pointers` compile and pass (static_assert pins the const contract)
- [x] Existing `query_result_test`, `precondition_test`, and all `*_grove_test` suites still pass

## Summary by CodeRabbit

* **Breaking Changes**
  * Query results now provide read-only access to keys instead of mutable access, enforcing const-correctness across operations.
  * Type constraints added to validate key types at compile time.

[Review Change Stack](https://app.coderabbit.ai/change-stack/genogrove/genogrove/pull/397?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


riasc:
## Code Review

Tight, well-scoped tightening of two related public types. The diff stays inside the two result headers + their test; no plumbing changes needed inside the grove because internal call sites already pass non-const \`key*\` that converts implicitly to \`const key*\` at the API boundary.

### Bug

None.

### Guideline violation

None.

### Suggestion (optional, skipped)

A runtime test that actually queries a grove and confirms iteration still works would be redundant — every \`*_grove_test\` already exercises that path and would have caught any breakage. The added \`static_assert\` tests pin the exact contract this PR is about (the const drift), which is the right tool for the job.

### Comments processed

- **CodeRabbit issue-level**: rate-limited again — no actionable feedback. The earlier rate-limit pattern (5 PRs back-to-back) seems to have continued; might be worth checking the CodeRabbit plan / quota before the next batch.
- No inline review comments.

### CHANGELOG

Added in 3dffcdf under \`### Changed\` with both \`[#324]\` and \`[#397]\` links. Entry calls out the source-level breaking aspect explicitly so users searching CHANGELOG for a compile error can find the explanation.

### CI

7/7 Debug jobs green on \`ab878b4\` (pre-CHANGELOG). Re-running now on \`3dffcdf\` — CHANGELOG-only commit, same expected outcome.

### Tier 2 status

Closes the first Tier 2 issue from the release-blocker survey. Last Tier 2 item is #332 (\`key::operator==\` requires-clause too loose) — small cheap fix; one more PR.

PR #398, opened 2026-05-17
key<>: value-only equality and ordering operators (==, <, >)

### LINKED ISSUE #332: [High] key::operator== requires-clause doesn't constrain data_type equality

riasc:
### Changed
- **`key<>` comparisons compare value only**: `operator==` no longer touches `data` (was the only thing requiring `data_type` to be equality-comparable). New `operator<` and `operator>` are added on the wrapper and delegate to `value` for consistency — they are unconditionally available because `key_type_base` already requires `<` and `>` on `key_type`. Two keys at the same `value` with different `data` now compare equal and order identically — matching how the B+ tree itself reasons about identity (search and ordering use `value::operator<`, not data). Frees `data_type` from needing any comparison operators of its own. The original issue ([#332](https://github.com/genogrove/genogrove/issues/332)) proposed tightening the `operator==` requires-clause to require `data_type` equality-comparability; instead we removed the dependency on `data_type` equality entirely and extended the same design to `<` / `>`. ([#332](https://github.com/genogrove/genogrove/issues/332))

Closes #332

## Impact

- **Semantic**: `key<X, Y>{v, d1} == key<X, Y>{v, d2}` now returns `true` (was `false`); `<` and `>` on `key<>` are new and identically value-only. These are the intentional changes.
- **Compile-time**: `key<X, Y>` is now equality-comparable as long as `X` is — `Y` no longer needs `operator==`. `<` and `>` were not previously defined on `key<>` at all; they delegate to `key_type` and inherit any constraints from the concept.
- **Verified in-tree call sites**: no production code, CLI handler, or test compares `key == key` and depends on data being part of equality. The B+ tree orders by value via `operator<` and identifies nodes by pointer.

## Tests

Uses the project's existing `TYPED_TEST_P` pattern (mirroring `tests/structure/key_type_grove_test.hpp`) so the contract is verified once per `key_type`:

- `key_comparison_typed_test::equality_compares_value_only` — runtime test pinning that same-value-different-data compares equal.
- `key_comparison_typed_test::equality_available_for_non_comparable_data` — compile-time `static_assert` that `std::equality_comparable<key<KeyType, opaque>>` holds for data with no `operator==`.
- `key_comparison_typed_test::ordering_compares_value_only` — runtime test pinning `<` and `>` delegating to value.

Instantiated for `interval`, `numeric`, and `genomic_coordinate` from each existing `<type>_key_test.cpp`. New shared header `key_comparison_typed_test.hpp` next to the per-type files.

## QC
- [x] New typed-test suite passes for all three instantiations (`Interval/key_comparison_typed_test.*`, `Numeric/key_comparison_typed_test.*`, `GenomicCoordinate/key_comparison_typed_test.*`)
- [x] Existing `IntervalKeyTest`, `NumericKeyTest`, `GenomicCoordinateKeyTest`, and all `*_grove_test` suites still pass

## Summary by CodeRabbit

* **Behavior Changes**
  * Key equality now compares only the key value, ignoring associated data types
  * Keys with non-comparable data can now satisfy equality checks
  * New ordering operators added for consistent key comparison

* **Tests**
  * Consolidated key comparison tests into a shared test suite for improved coverage

[Review Change Stack](https://app.coderabbit.ai/change-stack/genogrove/genogrove/pull/398?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


riasc:
## Code Review

CI is fully green (7/7 Debug jobs across macOS + Ubuntu × Clang 15–18 / GCC 13–14) and the diff is small and focused. No bugs, no project-convention violations.

### Findings

**Bug** — none.

**Guideline violation** — none.

**Suggestion (optional)**
- C++20 auto-generates `operator!=` from `operator==`, so `key<>` gets `!=` for free. It does **not** synthesize `<=` / `>=` from `<` / `>`, so `key<>` currently has `==`, `!=`, `<`, `>` but no `<=` / `>=`. Callers can write `!(a > b)` and `!(a < b)`, but if you want a complete relational set, the cleanest path is a defaulted `operator<=>` returning `std::strong_ordering` (or `auto` deduced from the inner type). The `key_type_base` concept only requires `<`, `>`, `==`, so the spaceship version would either need to assume more from `key_type` or fall back to two separate `<=` / `>=` overloads. Not a blocker for #332; worth a follow-up issue if you decide it matters.

### Comment assessments

Only one comment on the PR (a CodeRabbit rate-limit notice) and it carries no actionable feedback. No inline review comments. The CodeRabbit auto-summary in the PR body is accurate.

### Housekeeping done in this review
- Title refreshed to `key<>: value-only equality and ordering operators (==, <, >)` (the previous title only mentioned `operator==`).
- PR body summary, impact, tests, and QC sections rewritten to cover the added `<` / `>` and the renamed `key_comparison_*` test names.
- CHANGELOG `[Unreleased] / ### Changed` entry added in [50266da](https://github.com/genogrove/genogrove/commit/50266da).
- QC items 2–4 marked based on green CI. The human-check item remains unchecked.
- Label `enhancement` is correct.

Issue #350, opened 2026-05-16
[Medium] precondition_test missing static_asserts for zlib streambuf move-deleted


riasc:
## Finding

`tests/structure/precondition_test.cpp` — CLAUDE.md / project memory records that `detail::deflate_streambuf` / `inflate_streambuf` are copy AND move deleted (Rule of Five). The deletions exist (zlib_streambuf.hpp:39-42, 119-122) but no static_assert pins them.

## Risk

Easy to silently re-enable a move in a future refactor.

## Fix

Add to `ruleOfFiveTest`: `static_assert(!std::is_move_constructible_v<gst::detail::deflate_streambuf>);` plus the four other deleted-trait combinations. May require lifting `detail::` visibility to the test or testing through a wrapper.

## Source

Spotted during `/cpp-audit` on 2026-05-16. Related: #269

PR #402, opened 2026-05-18
registry: declare ~registry() = default for Rule of Five intent

### LINKED ISSUE #358: [Medium] registry: no explicit destructor declaration — Rule of Five intent unclear
## Finding

`include/genogrove/data_type/registry.hpp:107-110` — Copy and move are deleted, no destructor declaration.

## Risk

Singleton works fine without one, but Rule-of-Five-aware readers expect `~registry() = default;` to confirm intent.

## Fix

Add `~registry() = default;` next to the deleted copy/move declarations.

## Source

riasc:
## Summary

- Adds `~registry() = default;` next to the four already-deleted copy/move declarations so the Rule of Five is complete and unambiguous on the type. No behavior change — the implicit compiler-generated destructor is identical.
- Updates the trailing-block comment so it explains the intent (singleton; copy/move deleted; destructor defaulted) rather than just "Disable copy and move".

Closes #358.

## Test plan

- [x] All registry tests continue to pass (pure declaration change; no behavior path touched)
- [x] Build with `BUILD_TESTING=ON` and run `ctest` — all `RegistryTest.*` and `KeyPayloadRegistryTest.*` cases pass


riasc:
## Code Review

CI green on all 7 build matrix jobs (gcc 13/14, clang 16/17/18, Apple clang 15/16). Diff is 6 lines net — pure declaration + comment change. CodeRabbit rate-limited at the time of writing, so this is a self-review only.

### Findings
None — the change does exactly what the issue asked. Declaration sits in the canonical Rule-of-Five block, ordering preserves the existing layout (copy-ctor delete, copy-assign delete, move-ctor delete, move-assign delete, then defaulted dtor), and the rewritten comment captures the intent without surprises.

### Verified
- **No behavior change**: `~registry() = default;` produces the same destructor the compiler would synthesize implicitly. Singleton holds only `std::deque<Payload>` + `std::unordered_map<Key, id_type>` + `std::mutex` — all trivially destructible / RAII. Defaulted is correct.
- **Documents the singleton contract**: with all four copy/move members deleted plus an explicit defaulted dtor, the type is unambiguously non-copyable and non-movable to any reader scanning the special-member block.
- **No callers affected**: no public-API signatures changed; no internal call site touches the destructor.
- **Comment clarity**: parenthetical "the compiler-generated dtor is identical, so this is documentation, not behavior" preempts the obvious "why bother" reader question.

### Suggestion (optional, follow-up)
There's no existing test that *enforces* the deletions via `static_assert(!std::is_copy_constructible_v<registry<...>> && !std::is_move_constructible_v<registry<...>>)`. Issue #350 already tracks the same gap for the zlib streambufs ("precondition_test missing static_asserts for zlib streambuf move-deleted") — folding registry into that effort would be the natural place to add the equivalent assertions rather than expanding this PR. Not a blocker here.

CHANGELOG entry added on this branch under `[Unreleased] / ### Refactored` in a follow-up commit.

PR #414, opened 2026-05-20
io: explicit special members on reader bases and kstring_guard (#326, #375)

### LINKED ISSUE #326: [High] file_reader move special members rely on base implicit copy-assign — fragile
## Finding

`include/genogrove/io/file_reader.hpp:18-22`; `include/genogrove/io/bed_reader.hpp:117-135`; `include/genogrove/io/gff_reader.hpp:107-125` — The reader move ctor / move assign invoke `file_reader<EntryType>::operator=(std::move(other))`, but `file_reader_base` only has an implicit copy-assign (its virtual destructor suppresses implicit moves).

## Risk

Works today only because the base has no data members — any future addition of base state (logger, stats counter) would silently be copied (not moved).

## Fix

Explicitly default move special members on both `file_reader_base` and `file_reader<T>` (and explicitly delete copy). Apply uniformly across bed/gff/bam/fasta readers.

## Source

Spotted during `/cpp-audit` on 2026-05-16. Related: #148

### LINKED ISSUE #375: [Low] kstring_guard doesn't explicitly delete/default move special members
## Finding

`include/genogrove/io/kstring_guard.hpp:14-22` — RAII guard deletes copy ops but not move ops; implicit-move generation is suppressed by the user-declared destructor, so moves fall back to deleted copy.

## Risk

Works, but Rule of Five intent is unclear.

## Fix

Add `kstring_guard(kstring_guard&&) = delete; kstring_guard& operator=(kstring_guard&&) = delete;`.

## Source

Spotted during `/cpp-audit` on 2026-05-16. Related: #135

riasc:
## Summary

Two `/cpp-audit` Rule-of-Five findings — making implicit/fragile special-member behavior explicit on RAII / polymorphic types. No behavior change.

### Refactored

- **#326** `[High]` — every concrete reader (`bed_reader` / `gff_reader` / `bam_reader` / `fasta_reader`) already declares a full Rule of Five, and their move ctor/assign call through `file_reader<EntryType>`'s special members. But **neither base** — `file_reader_base` nor `file_reader<T>` — declared its own move members: `file_reader_base`'s user-declared virtual destructor suppresses implicit move generation, so a move through the base subobject silently fell back to the base's *copy*-assignment. Harmless today because both bases are stateless, but fragile — any future base state (a logger, a stats counter) would be copied, not moved, with no diagnostic. Both bases now explicitly: default-construct, **delete copy** (a polymorphic base copy would slice), and **default move**; `file_reader<T>` also defaults its destructor. The four concrete readers already declare their special members explicitly, so they need no change — the gap was purely in the bases.
- **#375** `[Low]` — `kstring_guard` (an htslib `kstring_t` RAII guard) declared deleted copy ops but not move ops. The moves are already implicitly unavailable — the user-declared destructor suppresses them — so this is intent-only: explicit `= delete` on the move ctor/assign states the non-movable Rule of Five outright instead of leaving it to the suppression rules.

### Why bundled

Same topic — making special-member intent explicit on RAII / polymorphic types flagged by the same cpp-audit pass. One review.

### Behavior

No behavior change. `file_reader_base` / `file_reader<T>` are stateless, so defaulting their move members produces identical results to the previous (copy-based) path; `kstring_guard`'s moves were already unavailable. The four concrete readers are untouched.

Closes #326, closes #375.

## Test plan

- [x] `cmake --build build -DBUILD_TESTING=ON && ctest --output-on-failure` passes — in particular the reader move-semantics tests (move ctor / move assign for bed/gff/bam/fasta readers)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Refactored
+- **Explicit special members on the reader bases and `kstring_guard`**: `file_reader_base` and `file_reader<T>` never declared their own move members — `file_reader_base`'s virtual destructor suppressed implicit moves, so a derived reader's move silently fell through to the base's copy-assignment. Both bases now explicitly default-construct, delete copy (a polymorphic-base copy slices), and default move; `file_reader<T>` also defaults its destructor. `kstring_guard` gains explicit `= delete`d move members (already implicitly unavailable). No behavior change — the bases are stateless, so defaulted moves are identical to the prior copy path. ([#326](https://github.com/genogrove/genogrove/issues/326), [#375](https://github.com/genogrove/genogrove/issues/375), [#414](https://github.com/genogrove/genogrove/pull/414))
+
 ### Fixed
 - **`get_current_line()` contract documentation**: `bam_reader::get_current_line()`'s docstring said "Number of records read so far", implying records *yielded* by `read_next()` — but it counts every record *consumed* from the file, including records dropped by `should_skip()`. Corrected the docstring to match (behavior is unchanged: counting skipped records is what makes the "record N" in error messages point at the real file position). Also added the previously-absent docstring on `file_reader_base::get_current_line()` defining the contract — a 1-based input-consumed position, reader-specific unit (physical line for BED/GFF including comments/blanks, record index for BAM including filtered records), reflecting input consumed rather than entries yielded. ([#328](https://github.com/genogrove/genogrove/issues/328), [#379](https://github.com/genogrove/genogrove/issues/379), [#413](https://github.com/genogrove/genogrove/pull/413))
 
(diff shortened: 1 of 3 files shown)


riasc:
## Code Review

Reviewed the special-member declarations on the reader bases and `kstring_guard`.

### Findings

No bugs, no guideline violations, no suggestions.

### Notes

- **#326**: the diagnosis holds — `file_reader_base`'s `virtual ~file_reader_base() = default;` suppresses implicit move generation, so the bases had implicit copy-assign and no move. Defaulting move + deleting copy on both bases is the right fix. Because the bases are stateless, a move through the base is byte-identical to the previous copy path — zero behavior change, just correct intent and future-proofing.
- The four concrete readers were verified to already declare a full Rule of Five (`bed`/`gff`/`bam`/`fasta` — copy `= delete`, user-defined move, destructor), so they correctly need no change; the gap was purely the bases.
- `file_reader<T>` deleting copy explicitly does not affect `file_reader_iterator`, which holds a `file_reader<T>*` (pointer) and remains copyable — required for post-increment. CI green confirms.
- **#375**: `kstring_guard`'s moves were already implicitly unavailable; the explicit `= delete` is intent-only, correctly commented as such.
- CI green across all 7 platforms — compiles and the reader move-semantics tests pass.
- CodeRabbit is rate-limited after this session's volume, so no bot pass; re-trigger with `@coderabbitai review` if wanted.

### Housekeeping

- Added `refactor` label.
- Added a `### Refactored` CHANGELOG entry under `[Unreleased]` referencing #326, #375, #414.
- Marked the `cmake --build` QC item (CI-verified); left the human-check item.

PR #428, opened 2026-05-21
fix: reuse the promoted key in split_internal_node (#389)

### LINKED ISSUE #389: [Medium] split_internal_node orphans the promoted separator key in key_storage
## Finding

`include/genogrove/structure/grove/grove_insert.ipp:339-349` — when splitting an internal node, the code computes a fresh aggregate `parent_key` covering the left half's subtree range and allocates it into `key_storage` via `allocate_key()`. The original `child->get_keys()[mid]` — the would-be promoted separator — is then dropped from the tree by `child->get_keys().resize(mid)` without ever being reclaimed. Its slot stays in `key_storage` forever.

```cpp
key_type parent_separator = child->get_keys()[0]->get_value();
for (size_t i = 1; i <= static_cast<size_t>(mid); ++i) {
    parent_separator = key_type::aggregate(parent_separator,
        child->get_keys()[i]->get_value());
}
gdt::key<key_type, data_type> parent_key{parent_separator};
auto* parent_key_ptr = allocate_key(parent_key);   // new slot

new_child->get_keys().assign(child->get_keys().begin() + mid + 1, child->get_keys().end());
child->get_keys().resize(mid);   // child[mid] silently abandoned
```

## Risk

Every internal-node split leaks exactly one slot into `key_storage`. Insert-heavy workloads (especially bulk loads against large groves) accumulate dead slots even without any `remove_key()` calls. Confirmed by `GroveCompactTest.NoOpWhenNothingRemoved` (the test was originally written to assert `key_storage_size()` is unchanged after `compact()` when no removals happened — it fails because `compact()` reclaims these orphans).

## Fix options

- **(a) Reuse the existing slot** — instead of allocating a new aggregate key and dropping `child[mid]`, mutate `child[mid]->set_value(aggregated_range)` and promote that pointer. Zero new allocations per split.
- **(b) Keep the new aggregate, remove the old** — harder, since `std::deque` doesn't support middle erase; would need a per-slot live/dead flag and rely on `compact()` to reclaim, which defeats the point.

Option (a) is the natural fix. The aggregate range that the new `parent_key` carries is identical to what (mid+1) separator slots would have carried in aggregate, so reusing `child[mid]` and mutating its value via `set_value()` (already used by separator updates elsewhere in the codebase) preserves all invariants. Saves both the allocation and the orphan.

## Workaround

Calling `grove::compact()` (added in #388) reclaims these orphans along with `remove_key()` leaks. But that's a post-hoc reclaim — fixing the source is cheaper and removes the need to compact after a pure insert workload.

## Source

Surfaced while implementing `compact()` for #322 (see PR #388). The orphans showed up as `key_storage_size()` shrinking after `compact()` on a grove that had never had `remove_key()` called.

riasc:
## Summary

`split_internal_node` leaked one `key_storage` slot on **every internal-node split**. It computed a fresh aggregate separator key (`allocate_key()` → a new slot), then dropped `child->get_keys()[mid]` — the would-be promoted separator — via `child->get_keys().resize(mid)` without ever reclaiming it. Insert-heavy workloads (especially bulk loads on large groves) accumulated dead slots even with no `remove_key()` calls.

**Fix** — reuse `key[mid]` as the promoted separator (option (a) from the issue): overwrite its value with the aggregate range via `set_value()` and hand the existing pointer to the parent. No new allocation, no orphaned slot. The promoted separator carries the exact value a freshly allocated key would have, so all B+ tree invariants are preserved and the produced tree is identical.

- `grove_insert.ipp` — `split_internal_node` reuses `child->get_keys()[mid]` instead of `allocate_key()`.
- `grove_remove_test.cpp` — `GroveCompactTest.PreservesIndexedKeysWithoutRemoval` now asserts `key_storage_size()` is **invariant** across `compact()` on a no-removal grove (it builds a multi-level tree, so it exercises `split_internal_node`). The test previously skipped that assertion specifically to tolerate this leak — that stale comment is removed.

## Test plan

- [x] `grove_remove_test` passes, including the strengthened `PreservesIndexedKeysWithoutRemoval` (`key_storage_size()` no longer shrinks under `compact()` without removals)
- [x] Existing `GroveRemoveTest` / `GroveCompactTest` and bulk-insert suites pass — the promoted separator value is unchanged, so tree structure is identical
- [x] Full build passes across the CI compiler matrix

Closes #389

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where internal node splits were creating orphaned storage slots. The system now properly reuses existing key storage during splits, improving memory efficiency and preventing storage waste.

[Review Change Stack](https://app.coderabbit.ai/change-stack/genogrove/genogrove/pull/428?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


riasc:
## Code Review

Reviewed the `split_internal_node` key_storage leak fix. **No bugs or guideline violations** — recommend merge.

### Verified

- **Correctness of the reuse** — `key[mid]` is precisely the key a B+ tree split promotes; reusing its object/slot and overwriting `value` with the left-subtree aggregate yields the identical separator a freshly allocated key carried. genogrove separators route on the aggregate *value*, not key identity, so routing is unchanged.
- **`data` field** — `key[mid]` is an existing internal-node separator, always created via the single-arg `key` constructor (`data` default-constructed), so reusing it leaves `data` exactly as a fresh separator would have it. `set_value()` touches only `value`.
- **Pointer safety** — `key[mid]` lives in `key_storage` (`std::deque`, stable addresses); `set_value` mutates in place. After `resize(mid)` the pointer is removed from the child's vector and re-homed into the parent — no dangling, no double-ownership.
- **No counter drift** — `allocate_key()` never touched `leaf_key_count` (separators aren't leaf keys), so dropping the call leaves `leaf_key_count` / `indexed_vertex_count()` unaffected.
- **Regression test** — `PreservesIndexedKeysWithoutRemoval` now asserts `key_storage_size()` is invariant across `compact()`; that grove builds a multi-level tree (so it exercises `split_internal_node`), making the assertion fail pre-fix and pass post-fix.

No **Bug**, **Guideline violation**, or **Suggestion** items.

CI is green across the full matrix (Apple Clang 15/16, Clang 16/17/18, GCC 13/14). An independent Codex review is posted separately.

riasc:
## Codex Review

Automated independent review by OpenAI Codex (`codex review`, branch diff vs `main`).

The change reuses the promoted internal separator slot without altering B+ tree routing semantics, and the updated test expectation matches the intended removal of orphaned key-storage slots. I did not find any actionable correctness issues in the diff.

PR #431, opened 2026-05-22
perf: micro-opts in query_result ctor and fasta_index::fetch (#356, #366)

### LINKED ISSUE #356: [Medium] query_result ctor: pass-by-value query then copy-construct member — missing std::move
## Finding

`include/genogrove/data_type/query_result.hpp:63` — `query_result(key_type query) : query(query), keys{} {}` — pass-by-value then copy-construct.

## Risk

Should `std::move`. Harmless for interval/genomic_coordinate today; costs an extra copy for future heavy key types.

## Fix

`: query(std::move(query)), keys{} {}`

## Source

### LINKED ISSUE #366: [Low] fasta_index::fetch(name) does two index lookups (length, then bounded fetch)
## Finding

`src/io/fasta_index.cpp:80-86` — `fetch(name)` calls `sequence_length(name)` then `fetch(name, 0, len)`, which re-validates `faidx_has_seq` and re-bounds-checks.

## Risk

Two index lookups per fetch.

## Fix

Inline the no-arg variant to skip the redundant `has_seq` check (call `faidx_fetch_seq` directly with the computed length).

## Source

riasc:
## Summary

Two cpp-audit perf findings, bundled — part of the v0.24.4 patch.

- **#356** — `query_result`'s constructor took `key_t query` by value, then copy-constructed the `query` member from it. The by-value parameter is a temporary about to die, so it is now `std::move`d into the member.
- **#366** — `fetch(name)` called `sequence_length()` (one name resolution via `faidx_seq_len`) and then delegated to `fetch(name, 0, len)`, which re-resolved the *same* name via `faidx_has_seq`. The post-validation fetch logic (coordinate guards + `faidx_fetch_seq`) is extracted into a shared `fetch_region()` helper; `fetch(name)` now calls it directly, resolving the name once instead of twice. `fetch(name, start, end)` keeps its `faidx_has_seq` check for the direct-call path.

Behavior and error messages are unchanged — `fetch_region()` carries the exact guards and throw strings the inline code had.

## Test plan

- [x] Full build passes across the CI compiler matrix
- [x] `indexedfasta-test` passes — both `fetch` overloads, the unknown-sequence throws, and the invalid-region throw still behave identically (they exercise `fetch_region`)

Closes #356
Closes #366

## Summary by CodeRabbit

* **Refactor**
  * Reduced memory usage by storing query parameters more efficiently.
  * Centralized and simplified sequence extraction logic for clearer flow and maintainability.
  * Documentation updated to reflect these internal optimizations with no behavioral change.

[Review Change Stack](https://app.coderabbit.ai/change-stack/genogrove/genogrove/pull/431?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Refactored
+- **Perf micro-opts in `query_result` and `fasta_index`**: `query_result`'s constructor moves its by-value query parameter into the member instead of copying it; `fasta_index::fetch(name)` resolves the sequence name once (via `sequence_length`) rather than twice — the post-validation fetch logic is now shared through a `fetch_region` helper. No behavior change. ([#356](https://github.com/genogrove/genogrove/issues/356), [#366](https://github.com/genogrove/genogrove/issues/366), [#431](https://github.com/genogrove/genogrove/pull/431))
+
 ### Fixed
 - **`find_child_pos` rejects a broken tree invariant**: it returned `siblings.size()` (an out-of-range index) when a node was not among its parent's children; it now throws `std::runtime_error` instead of silently handing back a bad index. ([#374](https://github.com/genogrove/genogrove/issues/374), [#430](https://github.com/genogrove/genogrove/pull/430))
 - **`bam_reader` names the tag in a truncated-aux-data error**: `parse_tags` now reports the last-attempted tag, and `read_next`'s "Truncated auxiliary data at record N" error appends ", last tag: XX" so malformed BAM auxiliary data is diagnosable. ([#355](https://github.com/genogrove/genogrove/issues/355), [#430](https://github.com/genogrove/genogrove/pull/430))
--- a/include/genogrove/data_type/query_result.hpp
+++ b/include/genogrove/data_type/query_result.hpp
@@ -8,6 +8,7 @@
 
 // Standard
 #include <stdexcept>
+#include <utility>
 #include <vector>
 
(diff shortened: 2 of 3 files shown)


riasc:
## Code Review

Reviewed the two perf micro-opts. **No bugs or guideline violations** — recommend merge. Both are behavior-preserving.

### Verified

- **#356** — `query_result(key_t query) : query(std::move(query))` correctly moves the by-value parameter into the member (it's a temporary about to die). Safe for every key type — trivially-copyable keys simply fall back to a copy. `<utility>` was added for `std::move`.
- **#366** — `fetch_region` carries the exact guards (`start >= end`, coordinate-limit) and `faidx_fetch_seq` error handling the inline code had — same throw types and strings. `fetch(name, start, end)` keeps its `faidx_has_seq` check for the direct-call path; `fetch(name)` relies on `sequence_length()` having already resolved the name, so the redundant second lookup is gone. The helper is correctly scoped (anonymous namespace inside `genogrove::io`), matching the existing `.cpp`-local-helper pattern.
- All `indexedfasta-test` paths still hold — known / unknown sequence, whole-sequence fetch, invalid region — each reasoned through `fetch_region`.

No **Bug**, **Guideline violation**, or **Suggestion** items.

CI is green across the full matrix. An independent Codex review is posted separately.

riasc:
## Codex Review

Automated independent review by OpenAI Codex (`codex review`, branch diff vs `main`).

The changes are small refactors/documentation updates and do not appear to introduce behavioral regressions. The FASTA helper preserves the existing validation and fetch semantics while removing a redundant lookup, and the query_result constructor change is safe for normal key types.

PR #436, opened 2026-05-23
revert: non-const key pointers in query_result / flanking_query_result

### LINKED ISSUE #435: Revert query_result to non-const key pointers
## What's wrong

`query_result::get_keys()` returns `const std::vector<const key<...>*>&` — pointers are const-qualified so callers can't mutate a key via `result.get_keys()[i]->set_value(...)` and silently corrupt B+ tree ordering.

In practice this makes every "build a graph from query results" example require a `const_cast<key<...>*>` to feed the keys back into `add_edge` / `link_if`, which take non-const pointers. The docs PR at genogrove/docs#138 (with underlying issue genogrove/docs#135) ran into this — the examples become ugly enough that we'd rather revert the library.

## Why the const isn't actually buying much

The same ordering-corruption footgun is already reachable via the non-const pointer that `insert_data()` returns:

```cpp
auto* k = grove.insert_data("chr1", interval{100, 200}, data);
k->set_value(interval{900, 1000});   // corrupts B+ tree ordering, no const blocks this
```

So `query_result` being the only place that enforces immutability *by type* isn't really preventing the misuse — it just makes the idiomatic path uglier while leaving the underlying footgun open elsewhere. A real safety story would have to be at the `key<>` level (e.g., gate `set_value` behind a friend declaration, or hand out a separate ordering-immutable `key_view` from both `insert_data` and `query_result`), not at `query_result` alone.

## Proposed change

Revert `query_result::get_keys()` to return `const std::vector<key<...>*>&` (non-const pointers). Same for `flanking_query_result::get_predecessor()` / `get_successor()`.

That makes the docs examples in `add_edge`, `link_if`, and friends compile as-is, restores the pre-v0.24.x ergonomics, and doesn't lose any safety that wasn't already missing elsewhere.

## Related

- Docs PR: genogrove/docs#138
- Docs issue closed in favour of this one: genogrove/docs#135

riasc:
## Summary

Resolves #435.

`query_result::get_keys()` and `flanking_query_result::get_predecessor()` / `get_successor()` are reverted to return non-`const` `key*` pointers so callers can feed result keys straight back into mutating graph operations (`add_edge`, `link_if`, …) without a `const_cast`.

The const-pointer variant from #324 / #397 only protected one of several entry points to the same ordering-corruption footgun — `insert_data()` already hands out a mutable pointer — while making the idiomatic call sites uglier (see the example that motivated this in genogrove/docs#138 / genogrove/docs#135). Restoring pre-v0.24.x ergonomics is the right trade.

## Changes

- `query_result`: stored vector and `get_keys()` return type back to `std::vector<key<...>*>` (non-const pointers); `add_key()` takes `key*`.
- `flanking_query_result`: predecessor/successor fields, getters, and setters back to `key*`.
- `query_result_test`: pin static_asserts flipped to enforce the non-const contract; the docstring is rewritten around #435's rationale.

Internal call sites (`grove_query.ipp`, `grove_flanking.ipp`) already pass non-const `key*`, so no caller-side changes were needed inside the library.

## Test plan

- [x] CI green (full ctest suite)
- [x] `tests/data_type/query_result_test.cpp` static_asserts compile and pass — pins the new return types
- [x] Existing query/flanking tests still pass (`grove_flanking_test`, `kmer_grove_test`, `genomic_coordinate_grove_test`, `heterogeneous_grove_test`, `graph_overlay_test`, `grove_remove_test`, `serialization_test`, `precondition_test`, `key_type_grove_test`)
- [x] CLI handlers (`gff`, `bed`) still compile against the new `get_keys()` signature

## Follow-ups

- Release PR bumping to v0.24.5 after this merges (covers `CMakeLists.txt` + `CHANGELOG` date-stamp + tag + GitHub release + docs version-bump issue)
- Docs PR genogrove/docs#138 can now be simplified to drop the `const_cast` workaround once v0.24.5 ships

## Refactor
* Query result accessor methods now return non-const key pointers instead of const pointers, enabling direct modification of keys in results.
* Affected methods: `get_keys()`, `get_predecessor()`, and `get_successor()`.
* Setter methods updated to accept non-const pointers.
* Improves API flexibility by eliminating the need for const_cast workarounds.
* Tests updated to verify the new behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/genogrove/genogrove/pull/436?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


riasc:
## Codex Review

_Posted by `/review-pr` — independent automated review run by OpenAI Codex against the branch diff vs `main`. Findings are verbatim from Codex._

Target: branch diff against `main`.

The patch introduces a const-correctness hole in the const flanking API that allows mutation of keys through a const grove and can corrupt tree invariants.

Review comment:

- **[P2] Preserve constness for flanking results** — `include/genogrove/data_type/flanking_query_result.hpp:58`
  Because `grove::flanking(...)` is a `const` method, returning a mutable `key*` here lets callers mutate keys through a `const grove` (for example `const grove& g; g.flanking(...).get_predecessor()->set_value(...)`). That can corrupt the B+ tree ordering while bypassing the container's non-const API; consider keeping const-returning access for results produced by const flanking queries or adding separate const/non-const overloads.

riasc:
## Code Review

Reviewed the PR against the issue, project conventions, and existing comments. The diff scope matches #435 exactly: pointer constness reverted on three files, internal callers were already non-const so no caller-side fixes were needed, and the pin static_asserts in `query_result_test.cpp` are correctly inverted to enforce the new contract.

CI: 7/7 matrix jobs green on the previous commit (gcc-13/14, clang-16/17/18, apple-clang-15/16); pending re-run on the CHANGELOG + trailing-newline follow-up commit `01b5c89`, which only adds whitespace to the C++ files and a CHANGELOG entry.

### Findings

#### Suggestion — `flanking()` is `const`, but now returns mutable `key*` (optional)

Codex (posted as a separate comment) flagged that `grove::flanking()` is declared `const`, yet `flanking_query_result::get_predecessor()` / `get_successor()` now return non-const `key*`. So a caller holding `const grove& g` can do `g.flanking(...).get_predecessor()->set_value(...)` and corrupt B+ tree ordering through a read-only handle.

This is a real const-correctness asymmetry — `grove::intersect()` is non-const, so `query_result` returning non-const `key*` is internally consistent; `grove::flanking()` is the one outlier. Three options:

1. **Accept the trade as-is.** The issue's reasoning ("the footgun is already reachable via `insert_data()` so restricting it at one boundary buys nothing") applies here too — and the v0.24.x docs PR genogrove/docs#138 ergonomics regression was the explicit motivation for the revert. This is what the PR currently does.
2. **Make `grove::flanking()` non-const.** Removes the asymmetry cleanly (matches `intersect()`), at the cost of `const grove&` consumers losing read-only flanking access.
3. **Two `flanking()` overloads** — non-const returning `key*`, const returning a `const`-keyed view. Restores both ergonomics and const-correctness, but adds API surface.

Per the issue's stated reasoning I read this as a known/accepted trade. Flagging for your call; not blocking.

#### Note — label

No `revert` label in the project's label set; chose `refactor` (closest existing label).

### Comment assessment

No inline review comments to address. CodeRabbit posted only a walkthrough with no actionable findings ("No actionable comments were generated in the recent review"). Codex's finding is captured above and in its own comment.

PR #451, opened 2026-06-13
Add node-less grove_to_sif(ostream) overload

### LINKED ISSUE #448: Add public node-less grove_to_sif(ostream) / to_sif(path) overload
## Problem

`grove::grove_to_sif` is public but its only signature takes an internal `node*`:

\`\`\`cpp
void grove_to_sif(std::ostream& os, const node<key_type, data_type>* root) const;
\`\`\`

To call it a caller needs a root node pointer, but `root_nodes` (the `index → root` map) is **private** and `node` is not part of the public surface. So there is no usable way to export a whole grove to SIF from outside the class — it's effectively an internals-only debug helper today.

This blocks exposing SIF export in **pygenogrove**: binding the current signature would require leaking `node` into Python (a lifetime/mutation hazard) or reaching into private state.

## Proposed change

Add a node-less public overload that walks the grove's own roots, e.g.:

\`\`\`cpp
// Export the entire grove (all indexed trees) to SIF.
void grove_to_sif(std::ostream& os) const {
    for (const auto& [index, root] : root_nodes) grove_to_sif(os, root);
}

// Optional convenience:
void to_sif(const std::filesystem::path& path) const;  // opens an ofstream, delegates
\`\`\`

Keeps the SIF-walking logic in tested C++; the existing `node*` overload can stay (private or public) as the per-tree primitive.

## Why

Lets pygenogrove bind SIF export as a thin pass-through (mirroring `serialize()`/`deserialize()`), without exposing tree internals. Tracked on the Python side under genogrove/pygenogrove#1.

riasc:
### Added
- **Node-less `grove_to_sif(std::ostream&)` overload**: walks the grove's own `root_nodes` and writes every indexed tree to SIF, so callers can export a whole grove without access to the private roots map or the internal `node` type. The existing per-tree `grove_to_sif(os, root)` primitive is unchanged. Empty grove → empty output; index iteration order follows the hash map (not stable across runs), matching the existing per-root behavior.
- Extends the typed `grove_to_sif_output` test (asserts the node-less output matches the per-root output for a single-index grove) and adds `grove_to_sif_empty_grove`.

Kept the API stream-only — no `to_sif(path)` convenience — to stay consistent with `serialize()` / `deserialize()`, which are `std::ostream&` / `std::istream&` only. Easy to add later if a path overload is wanted.

Closes #448

## QC
- [x] `genogrove_structure_tests` builds and the `grove_to_sif_*` typed tests pass across all key types
- [x] Node-less `grove_to_sif(os)` output is byte-identical to the per-root overload for a single-index grove
- [x] An empty grove (no indices) produces empty SIF output

## Summary by CodeRabbit

* **New Features**
  * Added a new export method to serialize entire tree collections in a single operation, eliminating the need for per-tree exports.

* **Tests**
  * Added comprehensive test coverage for the new export feature, including validation for empty collections.


riasc:
## Code Review

No issues found. The node-less `grove_to_sif(std::ostream&)` overload delegates to the existing per-root primitive, is correctly `const`, and introduces no new ownership or lifetime concerns. The `[index, root]` structured binding leaves `index` unused, which matches the existing `serialize()` loop over `root_nodes` in the same file and compiles clean across all CI compilers (GCC 13/14, Clang 16-18, Apple Clang 15/16).

API consistency is good: keeping it stream-only (no `to_sif(path)`) mirrors `serialize()` / `deserialize()`. Test coverage is appropriate — the typed test asserts the node-less output is byte-identical to the per-root overload for a single-index grove, and `grove_to_sif_empty_grove` covers the no-indices case. Non-deterministic index ordering is correctly documented and doesn't affect the single-index test.

CodeRabbit was rate-limited and didn't produce a review this run.

riasc:
## Codex Review

_Independent automated review by OpenAI Codex._

Skipped this run — Codex usage limit reached (resets Jul 2). Re-run `/review-pr 451` after that to include it.

Issue #463, opened 2026-07-02
Random-access (region / mmap-backed) deserialization of serialized groves (.gg)


riasc:
## Summary

Add **random-access (partial / region) deserialization** of the serialized grove (`.gg`): read only the part of the grove a query touches, instead of deserializing the whole file. Ideally memory-mapped so traversal pages in only the nodes it visits.

## Motivation

Today `.gg` is a single zlib stream, so `Grove.deserialize` must read the **entire** grove — a whole-genome grove is seconds of load + GB of RAM on every open, even for a single-locus query.

Region-reading the raw GFF (#456) avoids that for *located* queries, but it can only reconstruct a grove from the records **overlapping** the region — so it loses any graph edge whose endpoints fall outside the region, and cross-chromosome edges entirely (fusions, trans-regulatory links). A random-access `.gg` is strictly more powerful: it holds the **full, pre-built, fully-connected grove** and pages in only the touched part — so multi-hop *connected* queries, including cross-chromosome traversal, work with random access, no whole-genome load, and no per-query re-assembly of edges.

So #456 solves "read a region of the raw data cheaply"; this issue solves "query a region of the *pre-built connected grove* cheaply" — the piece needed for connected queries at genome scale.

## Current state

- `.gg`: whole-file zlib stream; `serialize`/`deserialize` are all-or-nothing.
- Structure: per-chromosome B+ tree (`grove<genomic_coordinate, json>`) + a graph overlay (labelled edges).

## Proposal

An indexed, block-structured `.gg` (format-versioned) plus a partial read path. Two classic shapes:

1. **Block-compressed + internal offset index** — serialize per-chromosome subtrees (or fixed-size pages) as independently-addressable blocks, with a header mapping `seqid`/coordinate → block offset. `Grove.deserialize(path, region=...)` reads the header, seeks, and decompresses only the needed blocks. (BAM = bgzf + `.bai`, applied to the grove.)
2. **mmap-friendly page-aligned layout** — lay the B+ tree out in fixed-offset pages and `mmap` the file; the OS lazily pages in only the nodes a traversal touches (LMDB-style on-disk B-tree). Compression complicates mmap (uncompressed pages, or block-compressed with a decode cache).

Sketch:
```cpp
auto g = grove::open_mmap(path);            // or Grove.deserialize(path, region)
auto res = g.intersect(query, "chr7");      // pages in only chr7's touched nodes
```

### The crux: the graph overlay must be lazily addressable too

[...]

PR #465, opened 2026-07-03
Reuse zlib state across serialization blocks (perf)

riasc:
## Summary

Recovers the serialize/deserialize performance regression introduced by the 0.2 block format (genogrove/genogrove#464). The block format compressed/decompressed each node block with a **fresh zlib stream per block** — a `deflateInit`/`inflateInit` (which allocates zlib's large internal state) for every node. On many small blocks that per-block setup dominated, giving a **2–4.5×** slowdown vs the old whole-file stream. The tell was the order-dependence: worst at small tree order (most, tiniest blocks), ~2× at large order.

### Changed

- **`block_deflater` / `block_inflater`** (in `zlib_streambuf.hpp`): one zlib state reused across all blocks via `deflateReset` / `inflateReset` — no per-block `init`/`end`, so the ~256 KB deflate-state allocation happens once per `serialize`/`deserialize` instead of once per node.
- **`memory_streambuf`**: a non-owning read-only `streambuf` over an existing buffer, so a decompressed block is parsed with no intermediate `std::istringstream` copy.
- **`serialize`**: writes the directory, then compresses each block with the reused deflater and writes it length-prefixed as produced — drops the whole-payload `std::vector<std::string>` buffering.
- **`deserialize`**: reuses one inflater + two scratch buffers across all blocks; decompresses into a reused buffer and parses via `memory_streambuf`.

### Not changed

- **On-disk 0.2 format is byte-identical** — still `[len][independent zlib stream]` per block (`deflateReset` produces the same output as a fresh `deflateInit` for the same input/level). No format bump; existing 0.2 `.gg` files remain readable.
- The streaming `deflate_streambuf` / `inflate_streambuf` are retained (still used by other code paths and tests).

The residual ~2× floor at large order (independent blocks can't share a zlib dictionary) is the intended cost of random-access addressability; **node-packing** and **store-raw-if-smaller** remain knobs for the PR 2 (lazy read) benchmark pass.

## Verification

Ran a standalone round-trip (order 3, cross-chromosome edges + metadata, 1200-key distributed external blocks spanning 3 chunks, leaf chain across splits, empty grove) — all assertions pass. Affected TUs (`serialization_test`, `precondition_test` which uses the streambufs directly, `graph_overlay`, CLI `bed_entry`/`gff_entry` payloads) syntax-check clean.

## Test plan

- [x] `cmake -S . -B build -DBUILD_TESTING=ON && cmake --build build`
- [x] `cd build && ctest --output-on-failure`
- [ ] Confirm the benchmark recovers (`BUILD_BENCHMARKS=ON`, `BM_serialization_size` / `BM_deserialization`) — expect the largest gains at small order

Refs #463

## Summary by CodeRabbit

* **Performance**
  * Improved grove save/load efficiency by compressing and writing blocks sequentially rather than buffering all compressed payloads up front.
  * Reduced memory overhead by reusing compression/decompression state across blocks during serialization and deserialization.

* **Bug Fixes**
  * Enhanced handling of truncated or incomplete compressed blocks to prevent corrupted reads.
  * Improved reliability of stream-based block processing for more consistent import/export behavior.


riasc:
## Code Review

CI is green across the full matrix. CodeRabbit raised two **Trivial** nitpicks; verdicts below.

### CodeRabbit nitpicks

**1. `write_block` allocates a fresh `ostringstream` + `raw.str()` copy per block (grove_serialize.ipp:189) — Address.**
Correct and on-theme: the compressed buffer (`comp`) is hoisted, but the *uncompressed* side still constructs an `ostringstream` and materializes a `std::string` copy per block. Fix: hoist a single `ostringstream` out of the loop (reset via `raw.str(std::string())` / `raw.clear()`), and pass `raw.view()` (C++20) to `block_deflater::compress` to drop the copy. Verified `view()` compiles under C++20 (GCC 13+, Clang 16+, Apple Clang 15+).

**2. Cap `clen` before `comp_buf.resize()` (grove_serialize.ipp:379) — Skip (with rationale).**
The existing guard rejects `clen > streamsize::max()`, which removes the real UB (a negative signed-cast read count). A `clen` above that is impossible; a large-but-valid `clen` yields a clean `std::bad_alloc` that propagates out of `deserialize` — the node-cleanup guard frees any parsed nodes, so there is no corruption or leak, only a failed allocation on a malformed file. A fixed "sane max block size" would risk rejecting legitimate large blocks (a `gff_entry`-heavy node can be sizeable), and there is no established untrusted-`.gg` threat model for this library. If that firms up (e.g. a service deserializing uploaded `.gg`), the right fix is a configurable cap or a seekable-remaining-bytes check, not a magic constant — I would rather not invent an arbitrary limit here.

### My own review

No correctness issues. The rewrite was verified by a standalone round-trip (order 3 → many small blocks, cross-chromosome edges with metadata, 1200-key distributed external blocks, leaf chain across splits, empty grove) — all assertions pass. The on-disk 0.2 format is byte-identical (`deflateReset` yields the same output as a fresh `deflateInit`), so no format bump and existing `.gg` files stay readable. `memory_streambuf` is read-only/sequential, which matches the parser (no seeks). The `uInt` cast on `avail_in` caps a single block at 4 GB — a non-issue (a block is one node or 512 external keys).

### Conventions

`enhancement` label fits (improves existing serialization). Streaming `deflate_streambuf`/`inflate_streambuf` retained for their existing users/tests; new helpers are non-copyable/movable (own a `z_stream`).

I can apply nitpick 1 on your go-ahead.

PR #466, opened 2026-07-03
Silence registry_test unused-result warnings with std::ignore

riasc:
## Summary

`registry_test.cpp` discarded the `[[nodiscard]]` return of `registry::deserialize` at ten call sites, emitting `-Wunused-result` warnings on every build. CONTRIBUTING.md's nodiscard policy documents `deserialize()` as `[[nodiscard]]` (line 20) and prescribes `std::ignore = expr;` for intentionally discarding such a return in tests or setup code (line 26). This wraps those ten calls with `std::ignore =`, honoring the policy as written.

No production code change. (An earlier revision of this PR removed the attribute from `registry::deserialize`; that was reverted in favor of the documented `std::ignore` approach — `registry.hpp` is now identical to `main`.)

Test-only; no CHANGELOG entry (internal test hygiene, no API or behavior change).

## Test plan

- [x] `cmake -S . -B build -DBUILD_TESTING=ON && cmake --build build` (no more `-Wunused-result` warnings from `registry_test.cpp`)
- [x] `cd build && ctest --output-on-failure`


riasc:
## Code Review

CI green across the full matrix; CodeRabbit posted no findings.

### Review

The change is exactly right and minimal — remove a misapplied `[[nodiscard]]` and document why. I verified it is the **only** such case in `registry`: `intern` keeps `[[nodiscard]]` (the returned id is the must-check result), the getters (`get`/`contains`/`size`/`empty`/`find`) are pure and correctly annotated, and `deserialize` was the sole side-effecting method whose return (a reference to the mutated singleton) is incidental. `grove::deserialize` correctly retains the attribute — it returns a fresh grove by value, so ignoring it wastes the call; the added comment captures that contrast.

No new test is needed: this removes an attribute, existing `registry` serialize/deserialize round-trip tests still exercise the function, and the effect (a warning-free build of `registry_test.cpp`) is itself the verification. Confirmed locally with `-Wall -Wunused-result`.

**No bugs, guideline violations, or suggestions.**

### Conventions

`refactor` label fits (annotation cleanup, no behavior change). `### Refactored` CHANGELOG entry added referencing #466 (no issue exists — this surfaced from a build log). The branch was updated from `main` first so the entry sits alongside the current #463/#464/#465 history.

---

## Update — approach reversed (a5e5c8a)

A follow-up finding (relayed from CodeRabbit) was **valid and important**: CONTRIBUTING.md's nodiscard policy explicitly lists `deserialize() → [[nodiscard]]` (line 20) and prescribes `std::ignore = expr;` for intentionally discarding such a return in tests (line 26). My original review under-weighted the documented policy (it exists in CONTRIBUTING.md, not just as a general principle).

So this PR is reversed to honor the policy as written:
- **`registry.hpp` restored to `main`** — `[[nodiscard]]` stays on `registry::deserialize`.
- The ten discarded calls in `registry_test.cpp` are wrapped with `std::ignore =` (exactly what line 26 prescribes).
- CHANGELOG entry dropped; relabeled `refactor` → `test`; no production/API change.

Verified clean with `-Wall -Wunused-result`. Net diff is test-only.

PR #472, opened 2026-07-08
Per-edge metadata in idx --links (optional 3rd TSV column)

### LINKED ISSUE #442: Per-edge metadata in idx --links (optional 3rd TSV column)
## Context

`--links` in #424 sub-item 3 accepts a 2-column TSV (`nameA<TAB>nameB`) and adds plain directed edges to the grove's `graph_overlay`. The overlay itself supports **optional per-edge metadata**, but the CLI surface doesn't expose it yet.

## Proposed design

Extend the `--links` TSV with an optional 3rd column whose value is attached to the edge as metadata:

```
gene1	exon3	supports_isoform=ENST00001
gene1	enhancerA
gene2	exon7	hi_c_score=4.2
```

- 2-column rows produce edges without metadata (current v1 behavior).
- 3-column rows produce edges with metadata = the 3rd field as a string.
- Wider rows (4+ columns) → error with line number.

## Open questions

- **Metadata type:** `std::string` is the simplest and matches the most likely use case (key=value annotations, scores serialised as strings). If `graph_overlay` later supports typed metadata more strictly, this can be extended.
- **Empty metadata field** (`A<TAB>B<TAB>`) — treat as "no metadata" or as "empty-string metadata"? Probably reject — ambiguous formatting is better caught.
- **Quoting / escaping** for tab characters in metadata — defer until a use case actually surfaces; for now, raw string with no escaping.

## Tasks

- [ ] Extend the TSV parser to accept an optional 3rd column
- [ ] Pass the metadata string to `graph_overlay::add_edge(a, b, metadata)`
- [ ] E2E test: a `.gg` built with metadata-bearing rows round-trips with the metadata visible in `graph_overlay::get_edges()`
- [ ] Document the 3-column format

## Related

- #424 — parent issue (2-column `--links` for BED lands first as sub-item 3)

riasc:
## Summary

Closes #442.

### Added

The `idx --links` TSV gains an **optional 3rd column**, attached to the graph edge as metadata:

```
gene1	exon3	supports_isoform=ENST00001   # edge WITH metadata
gene1	enhancerA                            # edge, no metadata (unchanged)
gene2	exon7	hi_c_score=4.2
```

Existing 2-column links files keep working unchanged — the 3rd field is purely additive.

- **Parser**: `parse_links_tsv` now returns `link_row{source, target, optional<metadata>}` and accepts 2 or 3 columns. 1 column, 4+ columns, or any empty column (including a present-but-empty 3rd) throw with the line number — an empty metadata field is ambiguous, so it's rejected rather than guessed.
- **Apply**: `links::apply_to_grove` attaches the 3rd field via `graph_overlay::add_edge(a, b, metadata)`; 2-column rows use the plain `add_edge(a, b)`. The dedup key now includes the metadata, so the same `(source, target)` pair carrying two **different** metadata values is kept as two parallel edges, while exact-duplicate rows are still collapsed.

### Changed (format)

`graph_overlay` per-edge metadata requires a non-`void` edge type, so the CLI's groves move from `edge_data_type = void` to `std::string` across `idx`, `isec`, and `grove_view`. **This changes the `.gg` on-disk format — existing indexes must be regenerated** (no serialization back-compat, per project policy). Metadata is a raw string; no quoting/escaping (tabs delimit).

## Test plan

- [x] `cmake -S . -B build -DBUILD_TESTING=ON -DBUILD_CLI=ON && cmake --build build`
- [x] `cd build && ctest --output-on-failure` — CLI links unit + index e2e suites, including the new metadata tests

Verified locally: full CLI + all three CLI test files compile clean; a standalone harness confirms metadata apply + exact-duplicate dedup, metadata surviving a serialize→deserialize round-trip, and same-pair-distinct-metadata yielding two parallel edges.

Follow-up: docs update in `genogrove/docs` after merge (3-column `--links` format).

## Summary by CodeRabbit

* **New Features**
  * Link TSVs can now include optional third-column per-edge metadata, which is preserved through indexing and output.
  * Multiple edges between the same source/target now coexist when their metadata differs.
* **Bug Fixes**
  * Improved link-file validation: only 2 or 3 columns are accepted, empty fields (including an empty metadata column) are rejected, and errors report the exact line number.
  * More robust handling of common line endings when reading link data.
* **Breaking Changes**
  * Index/intermediate formats for edge metadata now require regeneration to pick up the new metadata-aware edge payload behavior.


riasc:
## Code Review

Reviewed for correctness (with focus on the `void`→`std::string` edge-type/format switch), API consistency, and test coverage. CodeRabbit reported no actionable comments; I found no bugs.

**Verified (the format-switch is the risky part):**
- **Serialize/deserialize round-trip** of a string-edged grove preserves edge metadata (standalone harness).
- **`grove_view` over metadata-bearing edges** — the path `isec --in-place` newly exposes. Block parsing stays aligned past the per-edge metadata bytes; `get_neighbors` returns the right edges including cross-chromosome. No misalignment.
- All grove-type sites (idx write, isec deserialize, grove_view open, build-from-target, handler signatures) are consistent — CI's existing `idx`→`isec` e2e round-trips would fail on any mismatch, and they pass.
- Parser: 3rd column captured; empty-metadata and 4-column rejected; dedup key includes metadata (same pair, distinct metadata → parallel edges).

**Suggestion (optional, test-only):** the `grove_view`-over-metadata-edges path (`isec --in-place` on a `--links` index) is verified only by my local harness — the CI in-place e2e test uses an edge-free `.gg`. Recommend adding an e2e that runs `isec --in-place -i <metadata-links.gg>` so a future regression in block/metadata alignment is caught in CI. Happy to add it.

riasc:
## Update — added coverage

Added an e2e test (`InPlaceQueriesIndexWithMetadataEdges`) covering the previously-uncovered path: `isec --in-place` on a `--links` index whose edges carry metadata. It builds the index, queries it in place, and asserts the correct overlaps — so a future block/metadata alignment regression in `grove_view` is now caught in CI, not just my local harness. Pushed in 8fa7db7.

PR #473, opened 2026-07-08
Cross-type interval queries in intersect (BED↔GFF)

### LINKED ISSUE #439: Cross-type interval queries in intersect (BED query against GFF index, and vice versa)
## Context

When GFF support landed in #424 sub-item 1, `intersect` was wired to require **matching** query and target/index types (BED query for a BED target, GFF query for a GFF target). At the data-structure level this is unnecessarily restrictive — `grove::intersect()` only consumes `(interval, index_name)` tuples; the file format the query came from is irrelevant. The **output** format is what depends on the target's payload type (since that's what's in the result keys).

So all four combinations are well-defined:

| Query | Target / Index | Output format |
|---|---|---|
| BED | BED | BED rows |
| BED | GFF | GFF rows |
| GFF | BED | BED rows |
| GFF | GFF | GFF rows |

## Why it isn't done today

The current handlers (`handlers::bed::grove_intersect`, `handlers::gff::grove_intersect`) couple two responsibilities into one function:

1. **Query iteration** — read query records → produce `(interval, index_name)` tuples (depends on **query file type**).
2. **Result printing** — turn a result key's payload into an output row (depends on **target payload type**).

Coupling those is what forces the query type to match the target type.

## Proposed design

Split each handler into two primitives:

```cpp
// In handlers/bed:
template <typename F>
void for_each_bed_query(const std::string& queryfile, F&& cb);
//   cb(gdt::interval, const std::string& chrom)

void print_bed_result(std::ostream& out, const gio::bed_entry& payload);

// In handlers/gff:
template <typename F>
void for_each_gff_query(const std::string& queryfile, F&& cb);
//   cb(gdt::interval, const std::string& seqid)

void print_gff_result(std::ostream& out, const gio::gff_entry& payload);
```

Then a generic intersect runner in `intersect.cpp`:

```cpp
template <typename Payload, typename PrintFn>
void run_intersect(ggs::grove<gdt::interval, Payload>& g,
                   const std::string& queryfile,
                   gio::filetype query_type,
                   std::ostream& out,
                   PrintFn print) {
    auto handle = [&](gdt::interval iv, const std::string& idx) {
        for (auto* r : g.intersect(iv, idx).get_keys()) {
            print(out, r->get_data());
        }
    };
    if (query_type == gio::filetype::BED) {
        handlers::bed::for_each_bed_query(queryfile, handle);
    } else /* GFF / GTF */ {
        handlers::gff::for_each_gff_query(queryfile, handle);
    }
}
```

`intersect::execute()` then dispatches on the **target payload type** to instantiate `run_intersect` with the right printer; the **query type** is free.

## Tasks

- [ ] Extract `for_each_bed_query` / `print_bed_result` from `handlers::bed::grove_intersect`
- [ ] Extract `for_each_gff_query` / `print_gff_result` from `handlers::gff::grove_intersect`
- [ ] Add the generic `run_intersect` template in `intersect.cpp` (or a small new header)
- [ ] Rewrite `intersect::execute()` dispatch so query type and target type are independent
- [ ] Drop the four "query type must match target/index" error paths added in the GFF-support PR
- [ ] Add E2E tests for the BED-query-on-GFF-index and GFF-query-on-BED-index combinations

## Related

- #424 (GFF support — parent issue; matched-type intersect lands first as item 1 of #424)
- Surfaces during the matched-type implementation; deferred deliberately to keep that PR focused.

riasc:
## Summary

Closes #439.

### Added

`intersect` now allows the **query file type and the target/index type to differ** — all four combinations work, and the output format follows the *target* payload type:

| Query | Target / Index | Output |
|---|---|---|
| BED | BED | BED rows |
| BED | GFF | GFF rows |
| GFF | BED | BED rows |
| GFF | GFF | GFF rows |

- Each handler's `grove_intersect` is split into a **query iterator** (`for_each_bed_query` / `for_each_gff_query`, chosen by query type) and a **result printer** (`print_bed_result` / `print_gff_result`, chosen by target payload type). A generic `run_intersect` in `intersect.cpp` composes them.
- The four "query file must be X" errors are dropped. A query that isn't BED/GFF/GTF is still rejected.

### Coordinate handling (the crux — thanks @riasc)

BED is 0-based half-open, GFF is 1-based inclusive, and the CLI was converting them into **different** interval spaces — so a naive cross-type overlap would be off by one at the boundaries. Both readers now normalize to a common **0-based-inclusive** space:

- BED: `interval(start, end - 1)` (unchanged).
- GFF: `interval(start - 1, end - 1)` (was `interval(start, end)`), at both insert and query.

So a GFF feature `[101, 500]` (1-based) and the equivalent BED feature `[100, 500)` both map to `interval(100, 499)` and overlap identically.

**Unchanged:** same-type query results, and the printed coordinates (output always emits the raw entry coords of the target payload). **Changed:** the GFF stored interval numbers shift by −1 → the `.gg` format changes for GFF payloads, so **existing GFF indexes must be regenerated** (no serialization back-compat).

## Test plan

- [x] `cmake -S . -B build -DBUILD_TESTING=ON -DBUILD_CLI=ON && cmake --build build`
- [x] `cd build && ctest --output-on-failure` — CLI intersect unit + e2e suites, including the new cross-type tests

Verified locally: full CLI + all three CLI test files compile clean; a standalone harness over the real test data confirms BED-query→GFF-target and GFF-query→BED-target produce the correct overlaps and output rows, and that GFF→0-based boundary alignment is exact (first base hits, last base hits, one-past misses).

Follow-up: docs update in `genogrove/docs` after merge (cross-type intersect + the GFF coordinate normalization / regenerate-indexes note).

## Summary by CodeRabbit

* **New Features**
  * `intersect` now supports more cross-format combinations, including BED queries against GFF targets and GFF queries against BED prebuilt indexes.
  * Output generation is unified so results are consistently formatted per input type.

* **Bug Fixes**
  * Coordinate normalization was adjusted for GFF-based handling to report overlaps using expected 0-based, inclusive positions.
  * Validation of query file formats is now centralized, improving behavior for valid cross-type intersections.

* **Tests**
  * Updated end-to-end and unit tests to cover cross-format success paths and verify tab-separated output.


riasc:
## Code Review

Reviewed for correctness — focus on the coordinate normalization and the query/target decoupling. CodeRabbit reported no actionable comments.

**Bug found and fixed during review:** iterating `grove.intersect(...).get_keys()` directly in `run_intersect` dangled — `get_keys()` returns a reference into the temporary `query_result`, destroyed before the range-for body. This segfaulted the same-type BED path in CI's Debug build (the first push failed 7/8). Fixed by binding `query_result` to a local (in `run_intersect` and the unit helper); confirmed with AddressSanitizer (`stack-use-after-scope` on the un-bound pattern, clean on the fix). CI is now 8/8 green.

**Verified:**
- Cross-type both directions produce correct overlaps and target-format output (harness over the real test data).
- GFF→0-based **boundary alignment is exact**: GFF `[101,500]` ≡ `interval(100,499)` — first base hits, last base hits, one-past misses.
- Same-type results and printed coordinates unchanged (all pre-existing same-type e2e tests pass).

**Suggestion (my finding, optional):** the new GFF `interval(start-1, end-1)` underflows `size_t` if a malformed GFF row has `start == 0` (invalid for 1-based GFF) → `interval(SIZE_MAX, …)`, corrupting tree ordering. The old `interval(start, end)` didn't underflow. `gff_reader` currently accepts `start == 0` (it only checks `start > end`). Recommend rejecting `start < 1` in `gff_reader` where coordinates are parsed — single point, spec-correct. Happy to add it here.

PR #476, opened 2026-07-09
Reject GFF start < 1 in gff_reader (fixes #474)

### LINKED ISSUE #474: GFF start==0 underflows to SIZE_MAX in CLI interval conversion
## Context

Cross-type intersect (#473) normalized GFF coordinates to a 0-based-inclusive interval space: the CLI converts a GFF record to `gdt::interval(entry.start - 1, entry.end - 1)` at both insert (`handlers::gff::grove_insert`) and query (`handlers::gff::for_each_gff_query`).

`entry.start` is a `size_t` holding the native 1-based GFF start. `gff_reader` validates `start > end` but **does not reject `start == 0`** (it accepts the digit string `"0"`). GFF is 1-based, so `start == 0` is malformed input.

## Bug

For a malformed GFF row with `start == 0`, `entry.start - 1` underflows to `SIZE_MAX`, producing `interval(SIZE_MAX, end-1)` — a garbage interval that corrupts B+ tree ordering (and could trigger huge internal ranges). The pre-#473 conversion `interval(start, end)` did not underflow.

## Fix

Reject `start < 1` in `gff_reader` where the coordinates are parsed (next to the existing `start > end` check) — single point, spec-correct (GFF columns 4/5 are 1-based, minimum 1). This also protects any other consumer of `gff_reader`, not just the CLI.

Surfaced during review of #473; deliberately deferred to keep that PR focused.

riasc:
## Summary

Closes #474.

### Fixed

`gff_reader` accepted a `start == 0` coordinate (it only validated `start > end`). GFF columns 4/5 are **1-based**, so `start == 0` is malformed — and since #473 it flows into the CLI's 0-based conversion `interval(start - 1, end - 1)`, where `start - 1` underflows `size_t` to `SIZE_MAX`, producing a garbage interval that corrupts B+ tree ordering.

Now `start < 1` is rejected in both coordinate-validation paths:
- the constructor's first-record validation (throws `std::runtime_error`),
- the per-record `parse_line` path (sets `error_message`, honoring `skip_invalid_lines`).

The error message names the line. `start == 1` (the minimum valid GFF start) is unaffected. Fixing it in the reader protects every `gff_reader` consumer, not just the CLI.

## Test plan

- [x] `cmake -S . -B build -DBUILD_TESTING=ON && cmake --build build`
- [x] `cd build && ctest --output-on-failure` — new `validationRejectsZeroStart` (constructor) and `readNextRejectsZeroStart` (per-record) GFF tests

Verified locally: a standalone harness confirms the constructor rejects a zero-start first record, `read_next` rejects a later zero-start record, and `start == 1` is still accepted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for GFF files so records with a start coordinate of `0` are now rejected.
  * Parsing now stops with an error when invalid 1-based coordinates are encountered, helping prevent malformed data from being accepted.
* **Tests**
  * Added coverage for files with invalid start coordinates in both initial file loading and later record parsing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **GFF `start` coordinate < 1 rejected**: `gff_reader` accepted `start == 0` (it only checked `start > end`), which flowed into the CLI's 0-based conversion `interval(start - 1, end - 1)` and underflowed `size_t` to `SIZE_MAX`, corrupting B+ tree ordering. `start < 1` is now rejected in both the constructor's first-record validation and the per-record `parse_line` path (GFF columns 4/5 are 1-based); the error message names the line. `start == 1` is unaffected. ([#474](https://github.com/genogrove/genogrove/issues/474), [#476](https://github.com/genogrove/genogrove/pull/476))
+
 ## [0.25.0] - 2026-07-08
 
 ### Added
--- a/src/io/gff_reader.cpp
+++ b/src/io/gff_reader.cpp
@@ -144,6 +144,9 @@ namespace genogrove::io {
             if (ec1 != std::errc{} || ec2 != std::errc{}) {
                 throw std::runtime_error("Invalid GFF coordinates (out of range) in " + fpath.string());
             }
+            if (start_num < 1) {
+                throw std::runtime_error("Invalid GFF coordinates (start must be >= 1, GFF is 1-based) in " + fpath.string());
+            }
             if (start_num > end_num) {
                 throw std::runtime_error("Invalid GFF coordinates (start > end) in " + fpath.string());
             }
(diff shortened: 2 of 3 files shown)


riasc:
## Code Review

Small, correct bug fix — no findings, no CodeRabbit comments.

**Verified:**
- `start < 1` on a `size_t` is exactly `start == 0`; only malformed zero-starts are rejected, `start == 1` (minimum valid) passes.
- Guard added in both coordinate paths — the constructor first-record validation (throws) and the per-record `parse_line` (sets `error_message`, so `skip_invalid_lines` still applies to iteration). Consistent with how the reader already handles `start > end`.
- No existing GFF test uses `start == 0`, so nothing regresses; the two new tests (`validationRejectsZeroStart`, `readNextRejectsZeroStart`) lock both paths.
- Fixing at the reader closes the underflow for every `gff_reader` consumer, not just the CLI — the right altitude.

CI 8/8 green. QC (CI + build/ctest) checked; human line-check left for you.

PR #478, opened 2026-07-09
Add VCF/BCF as an isec query format

riasc:
### Added
- VCF/BCF is now a valid `-q` query file for `isec`, searchable against BED or GFF targets and prebuilt indexes. `vcf_entry` maps to the CLI's canonical 0-based space exactly like BED (`interval(start, end - 1)`; `len(REF) >= 1` means a SNP becomes a single-base interval, no `size_t` underflow). VCF is **query-only** — never a target/index payload — so there is no `.gg` format change and the index side (`grove_view`, `--in-place`) is untouched.
- New query-only handler `cli/include/handlers/vcf.hpp` (`for_each_vcf_query`), a VCF dispatch branch in `run_intersect`, and a widened query-format guard.
- `.bcf` extension now maps to `filetype::VCF` so binary BCF queries are accepted (`vcf_reader` already handles both).

## QC
- [x] `CLIIntersectE2ETest.VcfQueryAgainstBedTarget/VcfQueryAgainstGffTarget/VcfQueryAgainstPrebuiltIndexInPlace` pass (Debug build — the `get_keys()` dangling footgun only surfaces unoptimized)
- [x] `filetypeDetectorTest.vcfExtension` and `bcfExtensionResolvesToVcf` pass
- [x] Existing BED/GFF intersect e2e tests still pass (cross-type dispatch unchanged)

## Summary by CodeRabbit

* **New Features**
  * Added support for using VCF and BCF files as intersection query inputs (via `intersect -q`).
  * VCF/BCF queries are supported with BED targets, GFF/GTF targets, and prebuilt BED indexes (`--in-place` supported).
  * Automatic recognition of `.vcf` and `.bcf` query files.
* **Tests**
  * Added end-to-end CLI coverage for VCF-based intersection queries across BED/GFF target combinations and BED indexes.
  * Extended file type detection tests for `.vcf` and `.bcf`.


riasc:
## Code Review

Reviewed the diff against project conventions. CI is green across the full Debug matrix (GCC 13/14, Clang 16/17/18, Apple Clang 15/16) and CodeRabbit passed with no findings.

**Correctness** — no issues. `for_each_vcf_query` maps `vcf_entry` to the CLI's canonical 0-based space identically to BED (`interval(start, end - 1)`). Since `vcf_entry.end = pos + len(REF)` and VCF requires `len(REF) >= 1`, `end - 1 >= start` always holds — no `size_t` underflow, and a SNP correctly becomes a single-base interval. The query/payload dispatch is orthogonal (the #473 design), so the untouched index side (`grove_view`, `--in-place`) works with a VCF query for free — the `VcfQueryAgainstPrebuiltIndexInPlace` test confirms it.

**Guideline compliance** — clean. Handler mirrors the bed/gff query-only split; no grove/output coupling; SPDX header present. (Fixed one thing during review: `query.vcf` was missing its trailing newline — corrected in the follow-up commit.)

**Test coverage** — adequate. VCF query against both BED and GFF targets (cross-type output), plus `--in-place`; `.vcf` and `.bcf` both asserted to resolve to `filetype::VCF` using real fixtures.

**Suggestion (optional, out of scope)** — Symbolic/structural variants (`<DEL>`, `<INS>`, breakends) carry their span in `INFO/END`, not `REF`, so `vcf_entry.end = pos + len(REF)` maps them to a single-base interval — a `<DEL>` query won't span its true deletion length. This is inherited `vcf_reader` behavior, not introduced here, and role 1 is deliberately scoped to positional overlap. Worth a docs note and a potential follow-up (honor `INFO/END`) if SV workflows matter.

**Label** — `feature` is correct (new query format). No change.

**CHANGELOG** — added an `### Added` entry under `[Unreleased]` linking [#478]. No backing issue exists (conversational request), so PR-link only.

PR #483, opened 2026-07-15
Add grove_view::flanking via resolver-driven descent (#482)

### LINKED ISSUE #482: grove_view: support flanking() via a resolver-driven descent
Downstream (pygenogrove#61) wants `flanking()` on `grove_view` so genome-wide nearest-neighbour queries can page in only the blocks on the descent path, instead of falling back to a full `deserialize()` or a region-restricted build.

## Current state
`grove_view` covers the read path — `intersect`, `get_neighbors`, `get_neighbors_if`, `get_edges` — but has **no `flanking`**.

`intersect` works on a view because the overlap search was extracted into a **resolver-driven** free function (`detail::search_overlaps` in `query_engine.hpp`): the view drives it with its lazy `block_resolver`, so children/next-leaf references resolve to blocks paged in on demand.

`flanking` was never given that treatment. `flanking_descend` (`grove_flanking.ipp`) is a member of `grove` that walks `current->get_children()` and `child->calc_subtree_range()` through **raw materialized pointers**. On a view those children are block-ids that may never have been loaded, so the descent would dereference blocks that arent there.

## Request
Mirror what was done for `search_overlaps`:

1. Extend the resolver concept beyond `child(n, i)` / `next(n)` with what flanking needs — multi-child iteration and per-child aggregate range (`children(n)` / `subtree_range(child)`), or equivalent — keeping the pruning logic (`flanking_could_descend`) backend-agnostic.
2. Extract `flanking_descend` into a resolver-driven free function alongside `search_overlaps`.
3. Add `grove_view::flanking(query, index)` and the predicate-filtered `flanking(query, index, is_compatible)` overload, driven by the views `block_resolver`, with the same block-paging semantics as `intersect`.

## Notes
- Query-only, like the rest of the view surface (no insert/serialize).
- Interval-nesting predecessor rule (largest-end, not sort-order max) and the successor rule must be preserved — theyre the reason the descent cant just be swapped for the overlap engine.
- Once this lands, pygenogrove#61 is a ~20-line binding.

riasc:
Closes #482.

Downstream (pygenogrove#61) wants `flanking()` on `grove_view` so genome-wide nearest-neighbour queries page in only the blocks on the descent path, instead of falling back to a full `deserialize()`. This mirrors what was already done for `intersect()` / `search_overlaps`.

## What changed
- **`query_engine.hpp`** — new resolver-driven free functions in `detail`:
  - `search_flanking` — the single flanking descent (leaf candidate scan + internal pruning), extracted from `grove`.
  - `subtree_range` — resolver-driven analogue of `node::calc_subtree_range` for the last child's catch-all range (walks the right spine through the resolver).
  - `flanking_could_descend` — the backend-agnostic pruning predicate.
- **`grove_flanking.ipp`** — `flanking()` now delegates to `search_flanking` with the eager resolver; ~130 lines of member descent/pruning removed.
- **`grove_view.hpp`** — `flanking(query, index)` and the predicate-filtered `flanking(query, index, is_compatible)` overload, driven by the view's `block_resolver`.

## Notes
- **No resolver-concept extension needed.** Flanking consults only `child(n, i)`; child count follows from the B+ tree invariant (`keys.size() + 1`), and the last child's subtree range is computed via `subtree_range` through the same `child()` hook. Both `eager_resolver` and `block_resolver` already satisfy `overlap_resolver`.
- Interval-nesting **predecessor rule** (largest-end, not sort-order max) and the **successor rule** (smallest-start) are preserved — they live unchanged in `search_flanking`.
- Query-only; no insert/serialize surface added, no format change.

## Verification
Built a standalone eager-vs-view harness on real serialized `.gg` files:
- interval (incl. nested `[300,400]` vs `[320,330]` largest-end case), numeric scalar path, and the strand-predicate overload — all match the eager grove across a spread of queries.
- Narrow flanking query on a 500-key tree pages in **67 / 251** blocks (genuinely partial).

## Test plan
- [x] `MatchesEagerFlanking` — interval, incl. nested-interval predecessor rule
- [x] `FlankingPartialLoad` — narrow query loads a subset of blocks
- [x] `FlankingPredicateOverloadMatchesEager` — genomic_coordinate strand filter
- [x] `FlankingNumericMatchesEager` — scalar path

## Summary by CodeRabbit

* **New Features**
  * Added flanking nearest-neighbor queries to grove views, returning predecessor and successor matches.
  * Introduced overloads with an optional compatibility predicate for leaf candidates.
  * Supports interval, genomic coordinate, and numeric query types, including partially loaded data; missing indexes return empty results.
* **Bug Fixes**
  * Improved flanking traversal to correctly handle nested/overlapping ranges and preserve nearest predecessor/successor selection rules.
* **Tests**
  * Added parity and partial-loading assertions covering predicate filtering and multiple data types.
* **Documentation**
  * Updated the changelog with the new flanking capabilities.


riasc:
## Code Review

Reviewed the diff for correctness, convention compliance, API consistency, and test coverage. CI is green on all 7 configs (clang 15–18, gcc 13–14, apple-clang 15/16).

**Correctness — clean.** The flanking descent was moved verbatim into `detail::search_flanking`; the predecessor (largest-end) and successor (smallest-start) rules and the interval/scalar pruning are unchanged. I confirmed eager-vs-view parity out of band with a standalone harness on real serialized `.gg` files (interval incl. the nested `[300,400]`/`[320,330]` case, numeric scalar path, strand-predicate overload), and a narrow query on a 500-key tree pages in 67/251 blocks — genuinely partial. The existing eager `grove_flanking` tests still pass in CI, so the extraction is behavior-preserving.

**API consistency — good.** Reusing `overlap_resolver` (flanking only needs `child()`) rather than extending the concept is the right call; `subtree_range` is a faithful resolver-driven analogue of `node::calc_subtree_range`.

### Suggestion (optional) — constrain the `Pred` template parameter
`search_flanking` / both `flanking` overloads take an unconstrained `typename Pred`, while the sibling `get_neighbors_if` constrains its predicate with `std::predicate<Predicate, const edge_data_type&>`. A named concept applied at all three sites would match that convention and this project's concepts-throughout style:

```cpp
template<typename Pred, typename key_type>
concept flanking_predicate = std::predicate<Pred, const key_type&, const key_type&>;
```

Low-risk, and it moves a misuse from a deep template error to a clear constraint failure. This is the same point CodeRabbit raised in its nitpick — assessed as **valid / worth addressing**. (Its review is a summary comment with no inline thread to reply to, hence noted here.)

No bugs, no convention violations, no missing edge-case coverage found.

---

## Update — round 1

Applied the one actionable item in commit 389329e:

| Item | Status | Notes |
|------|--------|-------|
| Constrain flanking `Pred` (my suggestion / CodeRabbit nitpick) | ✅ Addressed | Added `detail::flanking_predicate` concept + `requires` on `search_flanking` and both `flanking(query, index, is_compatible)` overloads (389329e). |

No other items were open. Re-verified: eager-vs-view parity harness still passes and all three key-type instantiations compile with the constraint in place.

PR #495, opened 2026-07-15
Make the CLI fail cleanly on bad args and honor --version (#485)

### LINKED ISSUE #485: CLI aborts on invalid arguments; --version flag is dead
From `/cpp-audit`. CLI argument-handling robustness issues.

### Uncaught cxxopts exceptions → std::terminate (High)
- [ ] `cli/src/main.cpp:51, 78-79` — `options.parse(...)` and the subcommand `parse_args(...)`/`parse(...)` run **outside** the `try/catch` that starts at line 86, and neither `idx` nor `isec` calls `allow_unrecognised_options()`. A typo'd flag (`idx x.bed --bogus`), a non-integer `-k foo`, or a missing option value throws an uncaught `cxxopts::exceptions::*`, aborting via `std::terminate` (exit 134) with a raw message instead of a clean `Error: …` + exit 1.
  - **Fix**: move the parse calls inside a `try` translating to `std::runtime_error` → exit 1.

### `--version` is dead (Medium)
- [ ] `cli/src/main.cpp:45` — `-v,--version` is registered but `args.count("version")` is never checked; `genogrove --version` falls through to "No subcommand specified" → exit 1. A `config/version` facility already exists (`version_test.cpp`).
  - **Fix**: after the main parse, `if (args.count("version")) { print version; return 0; }`.

### Misleading / leftover help text (Low)
- [ ] `cli/src/intersect.cpp:69,71,78` — help strings copy-pasted from `idx`: `-q` says "to be indexed" (it's queried), `-t` says "BED file" (GFF/GTF also accepted), `-o` says "Write the index" (writes results).
- [ ] `cli/src/intersect.cpp:166,212` and `cli/src/index.cpp:105` — unused `compression` element in `auto [filetype, compression] = detect_filetype(...)`; use `auto [filetype, _] = …`.

riasc:
Closes #485.

The main-level and subcommand cxxopts parsing ran **outside** any `try/catch`, and the subcommands don't `allow_unrecognised_options`. So a typo'd flag, a non-integer value (`-k foo`), or a missing option argument threw an uncaught `cxxopts::exceptions::*` and aborted via `std::terminate` — the most common user-error path crashed instead of reporting cleanly. The `-v,--version` flag was registered but never checked, so `genogrove --version` fell through to "No subcommand specified" → exit 1.

## What changed
- **`cli/src/main.cpp`** — wrap all parsing + dispatch in the existing `try/catch (const std::exception&)`. cxxopts exceptions derive from `std::exception`, so they're now reported via `e.what()` and exit 1, alongside the domain errors from `validate`/`execute`. (Bare `e.what()` preserved — domain errors already self-prefix "Error: "; cxxopts messages are self-describing.)
- **`--version`** — print `genogrove <major>.<minor>.<patch>` from `genogrove/config/version.hpp` and exit 0.
- **`cli/src/intersect.cpp`** — correct the `-q`/`-t`/`-o` help strings (copy-pasted from `idx`: `-q` was "to be indexed", `-t` "BED file", `-o` "Write the index").
- **`idx`/`isec`** — name the unused `compression` element of the three `detect_filetype` structured bindings `_`.

## Verification
- Syntax-checked `main.cpp`, `intersect.cpp`, `index.cpp` against the real cxxopts/htslib headers.
- Added two e2e regressions (CI runs the real binary): `--version` prints a version + exit 0; a non-integer `-k` reports an error + exit **1** (previously `std::terminate` → non-normal exit).

## Test plan
- [x] `VersionFlagPrintsVersionAndExitsZero` — `--version` prints version, exit 0
- [x] `InvalidOptionValueFailsCleanly` — non-integer `-k` exits 1, no crash
- [x] Existing `idx`/`isec` e2e suites unaffected (help-text and binding changes are behavior-preserving)

## Summary by CodeRabbit

* **New Features**
  * Added `--version` support to display the installed Genogrove version.
  * Updated intersect command help text to clarify supported query and target file formats.

* **Bug Fixes**
  * Improved command-line error handling so invalid options exit cleanly with an error message instead of crashing.

* **Tests**
  * Added coverage for version output and invalid option handling.


riasc:
## Code Review

Reviewed for correctness, convention compliance, and coverage. CI is green on all 7 configs (clang 15–18, gcc 13–14, apple-clang 15/16).

**Correctness — clean.** All parsing + dispatch is now inside the existing `try/catch(std::exception)`; cxxopts exceptions derive from `std::exception`, so bad flags / non-integer values / missing arguments report via `e.what()` and exit 1 instead of `std::terminate`. Bare `e.what()` is preserved (domain errors already self-prefix "Error: "; cxxopts messages are self-describing). `--version` is checked right after the top-level parse and prints from `genogrove/config/version.hpp`. No correctness issues found.

**Coverage — good.** Two e2e regressions run against the real binary: `--version` → version + exit 0, and non-integer `-k` → clean exit 1 (previously a `std::terminate`, which `pclose` would surface as a non-normal exit).

No bugs, no guideline violations.

### CodeRabbit nitpicks — assessed
Both are in the review body (no inline threads), so noting verdicts here:
- **`_` unused structured-binding elements** (`index.cpp:105`, `intersect.cpp:166/212`) — **Address.** No warning fires today (no strict flags), but `_` is a real variable and the two in `intersect::execute` sit in nested scopes (theoretical `-Wshadow`/`-Wunused`). Switching to `auto filetype = std::get<0>(...)` drops the binding entirely and is cleaner.
- **Drop `argc/argv` from `subcall::parse_args`** — **Skip (out of scope).** The params are genuinely unused, but removing them changes the abstract `subcall` interface + both impls + the call site — an unrelated refactor that belongs in its own PR, not this arg-robustness fix.

---

## Update — round 1

Applied the one in-scope item in commit ea1e86f:

| Item | Status | Notes |
|------|--------|-------|
| `_` unused binding → `std::get<0>` (index.cpp:105, intersect.cpp:166/212) | ✅ Addressed | Drops the binding entirely; no shadow/unused-var risk (ea1e86f). |
| Drop `argc/argv` from `subcall::parse_args` | ⏭️ Not addressed | Out of scope — unrelated `subcall` interface refactor; belongs in its own PR. |

Re-verified: both handler TUs compile against the real cxxopts/htslib headers.

PR #500, opened 2026-07-15
Remove orphaned deflate_streambuf / inflate_streambuf (#489)

### LINKED ISSUE #489: Remove orphaned zlib_streambuf streaming compressor
From `/cpp-audit`.

`detail::deflate_streambuf` / `detail::inflate_streambuf` (`include/genogrove/structure/grove/zlib_streambuf.hpp`) have **no production callers** — grove serialization moved to `block_deflater`/`block_inflater` (format 0.2). Verified: the only references are two test files (`tests/structure/precondition_test.cpp:163-164`, `tests/structure/serialization_test.cpp:42`).

- [ ] The class-level docstring ("compressed in chunks to avoid buffering the entire grove in memory", ~line 24) is now stale and reads as if these still *are* the grove serializer.
- [ ] `deflate_streambuf::overflow`/`sync` (lines 56-68) throw on zlib/write errors; streambuf virtuals are conventionally non-throwing (they may run during stream flush/unwinding). Only latent because nothing attaches these to a live stream today.

### Fix
Delete both classes and their two tests if `block_inflater`'s seekable-concatenation contract fully covers the need; otherwise update the Doxygen to state they're a standalone utility no longer used by grove serialization (and make the virtuals return `eof()`/`-1` instead of throwing).

_(CLAUDE.md's "streaming deflate/inflate_streambuf" description is correspondingly stale, but that file is owner-managed — flagging, not touching.)_

riasc:
Closes #489.

Grove serialization moved to the one-shot block codecs (`block_deflater` / `block_inflater`, format 0.2). The streaming `deflate_streambuf` / `inflate_streambuf` classes in `zlib_streambuf.hpp` had **no production callers** — only two test references — plus a stale class docstring that read as if they were still the grove serializer, and the throwing streambuf virtuals the audit flagged.

## What changed
- **`zlib_streambuf.hpp`** — delete both streaming classes (182 lines) and their now-unused `<istream>`/`<ostream>` includes. Kept: `block_deflater`, `block_inflater`, `memory_streambuf` (the active serialization path).
- **`precondition_test.cpp`** — the deleted test pinned a z_stream double-free guard (non-copyable/non-movable `static_assert`s). That rationale applies **equally to the live `block_deflater`/`block_inflater`**, which also own a `z_stream` and are move-deleted — so I redirected the guard to them rather than dropping the coverage. (`zlibStreambufsAreNonCopyableNonMovable` → `zlibBlockCodecsAreNonCopyableNonMovable`.)
- **`serialization_test.cpp`** — reworded the `#301` regression comment so it no longer names the deleted `inflate_streambuf`; the test itself (deserialize leaves the underlying stream clean) is unchanged and still valid for format 0.2.

## Verification
- No remaining references to either deleted class anywhere in the tree.
- `grove.hpp` / `grove_view.hpp` compile; a serialize → `grove_view` read round-trip is unchanged (block codecs untouched).
- Both touched test files compile clean against the real gtest/htslib headers.

## Note
CLAUDE.md's module description still mentions "streaming `deflate_streambuf`/`inflate_streambuf`", which is now stale — flagging for you to update (that file is owner-managed, so I didn't touch it).

## Test plan
- [x] `ruleOfFiveTest.zlibBlockCodecsAreNonCopyableNonMovable` — z_stream double-free guard now on the live codecs
- [x] `SerializationTest.FileStreamStateCleanAfterDeserialize` — deserialize leaves stream clean (unchanged behavior)
- [x] Full serialization/grove_view suites unaffected (block codecs untouched)


riasc:
## Code Review

Reviewed for correctness and coverage. (CodeRabbit skipped this PR; CI is running.)

**Correctness — clean.** This is a pure dead-code removal: the deleted `deflate_streambuf`/`inflate_streambuf` had no production callers (verified — zero remaining references anywhere in the tree), and the active serialization path (`block_deflater`/`block_inflater`/`memory_streambuf`) is untouched. I confirmed a serialize → `grove_view` read round-trip still works and both touched test files compile.

**Coverage preserved, not dropped.** The nice call here is redirecting the precondition test's `z_stream` double-free guard onto the live block codecs instead of deleting it — those classes own a `z_stream` and are move-deleted, so the invariant still matters. The `#301` regression test keeps its clean-stream assertion; only the comment (which named the deleted class) was reworded.

**Includes.** Dropping `<istream>`/`<ostream>` is safe — the header's remaining classes don't use them, and consumers (`grove.hpp` via `<sstream>`/`pod_io.hpp`, `grove_view.hpp` directly) supply them independently; both compile.

No bugs, no guideline violations. One out-of-band note (already in the PR body): CLAUDE.md's module description still mentions the streaming streambufs and is now stale — left for the owner to update.

PR #501, opened 2026-07-16
Deduplicate overflow-cascade split logic across insert paths (fixes #490)

### LINKED ISSUE #490: Deduplicate overflow-cascade split logic across the three insert paths
From `/cpp-audit`.

The "node full → (parent null: new root + split + republish root) else split parent's last child" cascade is duplicated across three insert paths:

- [ ] `include/genogrove/structure/grove/grove_insert.ipp:106-131` (bulk-append)
- [ ] `include/genogrove/structure/grove/grove_insert.ipp:426-444` (`insert_sorted`)
- [ ] `include/genogrove/structure/grove/grove.hpp:198-206` (`insert`)

### Risk
This split/rebalance path is already subtle (the sorted-append asymmetric midpoint). A fix must land in three places; easy to miss one and let them diverge.

### Fix
Extract a private helper, e.g. `cascade_split_from(node* overflow_node, std::string_view index, bool sorted_append)`, and call it from all three sites.

riasc:
Closes #490.

`/cpp-audit` flagged the "node full → cascade split" block as duplicated across three insert paths and warned they could diverge. They already had: the two sorted-append copies were **not** identical, and one was **wrong**.

## The bug
The bulk sorted-append path (append into a non-empty index) handled leaf overflow with a single-level `if` — it split the overflowing leaf's parent but never re-checked whether that parent itself overflowed as a result. `insert_sorted()` handled the same situation with a `while`-loop that cascades all the way up.

Consequence: appending enough sorted keys into an already-populated index left one internal node growing without bound. Queries still returned every key (`intersect` scans all children regardless of count), so it never threw or segfaulted — but the B+ tree invariant "no node exceeds `order-1` keys" was violated and that node degraded to O(n). This only affects the **append-into-non-empty** path; the common empty-index bulk build (bottom-up) was always correct.

## What changed
- **`grove_insert.ipp`** — extract the correct `while`-cascade into a private `cascade_split_sorted_append()` and call it from **both** sorted-append sites (bulk-append loop + `insert_sorted()`), so they can no longer diverge. The bulk-append `if` is replaced by a call + rightmost refetch.
- **`grove_insert.ipp`** — extract the identical root-promotion step (wrap overflowing root in a fresh parent, split, publish) into `promote_new_root(old_root, index, sorted_append)`, shared by all **three** insert paths (the third is the unsorted recursive `insert()`, `sorted_append=false`). This addresses the third site the audit named without changing its recursive cascade.
- **`precondition_test.cpp`** — regression test: seed one key, bulk-append 499 more at `order 3`, then run the shared tree validator. Fails on the pre-fix header (`Node has too many keys at depth 0 (has 249, max 2)`), passes after.

## Verification
- New regression test passes on the fix, **fails** on the pre-fix header.
- Full `precondition_test` suite (26 tests) green.
- Standalone probes at `order 3`: bulk-append (2000 keys) and scrambled unsorted `insert()` (1500 keys) both produce fully balanced trees — every key findable, no over-full nodes, no bad child counts, all leaves same depth.

## Test plan
- [x] `groveEdgeCaseTest.bulkAppendIntoNonEmptyIndexCascadesParentSplits` — new regression test
- [x] Full `precondition_test` suite passes across the CI matrix

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved B+ tree insertion and bulk append handling when node overflows require multiple parent splits.
  * Ensured sorted insert operations consistently maintain tree structure and rightmost-node tracking.

* **Tests**
  * Added coverage for bulk appending to a non-empty index, including validation that all appended entries remain searchable.


riasc:
## Code Review

**Verdict: approve.** No bugs, no guideline issues.

- **Correctness** — the extracted \`cascade_split_sorted_append()\` is `insert_sorted()`'s already-correct \`while\`-cascade verbatim; the bulk-append path now calls it (was a single-level \`if\` that left a parent overflowing). The shared \`promote_new_root()\` is a byte-for-byte extraction of the three root-promotion blocks, parameterized only on \`sorted_append\`. Verified all three insert paths (unsorted \`insert()\`, \`insert_sorted()\`, bulk-append) build balanced trees at order 3 with 1500–2000 keys.
- **Regression test** — \`bulkAppendIntoNonEmptyIndexCascadesParentSplits\` fails on the pre-fix header (internal node with 249 keys, max 2) and passes after; full \`precondition_test\` suite green.
- **Scope note** — filed as a refactor (#490), but one of the three duplicated copies was genuinely broken, so this is labeled \`bug\`. The empty-index bulk build (bottom-up) was never affected.

CodeRabbit: no actionable comments.

_No project-convention deviations. CHANGELOG entry added under `### Fixed`._

PR #502, opened 2026-07-16
Deduplicate CLI idx/isec handler write + dispatch tails (fixes #491)

### LINKED ISSUE #491: Deduplicate CLI idx/isec handler write + dispatch tails
From `/cpp-audit`.

- [ ] `cli/src/index.cpp:130-179` — the BED and GFF/GTF branches of `index::execute` duplicate the whole open-output / write-header / `serialize` / post-write-check sequence (144-152 vs 170-178), differing only in payload type and name-map construction. Extract `write_index<payload_t>(grove, outputfile, payload_type)`.
- [ ] `cli/src/intersect.cpp:144-230` — `intersect::execute` is ~85 lines with a 4-way near-identical dispatch (BED eager / BED in-place / GFF eager / GFF in-place) plus a 2-way target dispatch, each block varying only the grove type and printer. Extract a `dispatch_by_payload` helper (or a payload_type → {grove type, printer} table) so the eager/in-place + run-intersect logic is written once.

riasc:
Closes #491.

`/cpp-audit` flagged two near-identical dispatch tails in the CLI. Both are pure, behaviour-preserving extractions.

## What changed
- **`cli/src/index.cpp`** — the BED and GFF branches of `index::execute` duplicated the entire open-output → write-header → `serialize` → post-write-check sequence, differing only in grove type and payload enum. Extracted `write_index<grove_t>(grove, outputfile, payload_type)` (file-local template) and call it from both branches.
- **`cli/src/intersect.cpp`** — the index path had a 4-way dispatch (BED/GFF × eager/in-place) where each block varied only the payload type and printer. Extracted `query_index<payload_t>(...)` which makes the `grove_view` (in-place) vs `deserialize` (eager) choice once. The index path is now two payload branches over one shared helper instead of four hand-written blocks.

The `-t` target path (build a grove in memory, then query) was left as-is: it has no eager/in-place split, so its two branches are already the minimal "construct + insert + run" and differ by per-handler `grove_insert`/printer — wrapping them would add callable machinery for no real dedup.

## Verification
- Both TUs compile clean.
- Built the CLI and ran a full round-trip on **BED and GFF**: `idx` (both payload types), then `isec` via `-i` (eager), `-i --in-place` (`query_index` grove_view path), and `-t` target — all produce identical, correct overlaps.
- Covered in CI by `tests/cli/index-e2e-test.cpp` and `intersect-e2e-test.cpp`.

## Test plan
- [x] `index-e2e-test` — idx writes valid .gg for BED and GFF (write_index)
- [x] `intersect-e2e-test` — isec eager + in-place for BED and GFF (query_index)
- [x] Full CLI suite passes across the CI matrix


riasc:
## Code Review

**Verdict: approve.** Behavior-preserving extraction, no bugs.

- **`write_index<grove_t>`** — exact extraction of the open/header/serialize/post-check tail; the pre-check (build grove before opening output, so a parse error never truncates an existing `.gg`) is preserved because the call sites still build first.
- **`query_index<payload_t>`** — collapses the 4-way BED/GFF × eager/in-place dispatch to two payload branches over one helper; the `grove_view` (in-place) vs `deserialize` (eager) split is written once. `in` is passed through unused on the in-place branch — harmless, needed for the eager branch.
- **Left alone** — the `-t` in-memory-target path (no eager/in-place split; wrapping it would add callable machinery for no dedup). Reasonable.
- **Verification** — built the CLI and ran `idx` + `isec` (eager, `--in-place`, `-t`) on both BED and GFF; identical correct overlaps. Covered in CI by `index-e2e-test` / `intersect-e2e-test`.

Note: CodeRabbit hit its per-developer rate limit and did **not** review this PR — so there is no bot pass here, only this review + the green CI matrix.

_No project-convention deviations. CHANGELOG entry added under `### Refactored`._

PR #507, opened 2026-07-22
Clarify BGZF-reports-GZIP in filetype_detector; fix stale bam_reader usage example (fixes #494)

### LINKED ISSUE #494: Minor io papercuts: filetype_detector labels BGZF as GZIP; stale bam_reader docstring
From `/cpp-audit`. Two low-severity io fixes.

- [ ] `src/io/filetype_detector.cpp:41-89` (`detect_compression` / `detect_filetype`) — BAM and BGZF `.vcf.gz` report `compression_type::GZIP` because BGZF shares the gzip `1f 8b` magic, so the tuple is e.g. `(BAM, GZIP)`. Suppress compression classification for inherently-BGZF-framed types (BAM/BCF), or document that GZIP here means "BGZF-or-gzip container," so callers don't double-decompress.
- [ ] `include/genogrove/io/bam_reader.hpp:311` — the class usage example calls `grove.insert_data(entry.chrom, entry.interval, entry)`, but `sam_entry` has no `interval` member (it exposes `start`/`end`/`consumes_reference()`). Update to `gdt::interval(entry.start, entry.end - 1)` guarded by `entry.consumes_reference()`, matching the vcf_reader.hpp example.

riasc:
### Changed

Two low-severity io papercuts from `/cpp-audit` (issue #494), both documentation/comment only — no behavior change.

**1. `filetype_detector` — BGZF is reported as GZIP.** BAM, BCF, and bgzipped `.vcf.gz` share gzip's `1f 8b` magic, so `detect_filetype` returns `compression_type::GZIP` for them (e.g. `(BAM, GZIP)`). Rather than suppress compression classification for these types — which is impossible to do cleanly for BCF (it maps to `filetype::VCF`, indistinguishable from a plain `.vcf` post-mapping) and would break the existing BAM test that pins `GZIP` — the semantics are documented so callers don't double-decompress:
- Expanded the `compression_type::GZIP` enum comment.
- Added a `detect_filetype` docstring: the compression component is the outer container; BGZF-framed formats report GZIP and the htslib-backed readers unwrap the framing themselves.
- Added a comment at the gzip-magic check in `detect_compression`.

No production caller consumes the compression component (the CLI uses `std::get<0>` only), so this is purely a clarity fix.

**2. `bam_reader` usage example was stale.** The class docstring example called `grove.insert_data(entry.chrom, entry.interval, entry)`, but `sam_entry` has no `interval` member (it exposes `start`/`end`/`consumes_reference()`). Corrected to `gdt::interval(entry.start, entry.end - 1)` guarded by `entry.consumes_reference()` — unmapped / zero-ref-consuming reads have `end == start` and would underflow — matching the `vcf_reader` example.

## Test plan

- [x] CI green (comment/docstring-only change; no code paths altered — existing filetype/bam tests unaffected)

## Summary by CodeRabbit

* **Documentation**
  * Updated BAM reader examples to safely handle records that do not consume reference sequence and to calculate reference intervals correctly.
  * Clarified that BGZF-framed files, including BAM, BCF, and bgzipped VCF files, are identified as GZIP based on shared file signatures.
  * Documented that these files must not be decompressed with an additional gzip step.


riasc:
## Code Review

Reviewed against project conventions. Documentation/comment-only — no code paths altered.

**No blocking findings.**

- **`filetype_detector`**: I confirmed the *documented* fix is the right call over suppressing compression classification. BCF maps to `filetype::VCF` (indistinguishable from a plain `.vcf` after mapping), so a "suppress GZIP for BCF" branch can't be written cleanly; and `bamfile-test.cpp:1059` already pins `(BAM, GZIP)`, which no production code consumes (the CLI reads `std::get<0>` only). The enum note + `detect_filetype` docstring + magic-check comment make the "GZIP means gzip-or-BGZF container, don't double-decompress" contract explicit without touching behavior or tests.
- **`bam_reader` example**: the old example referenced `sam_entry::interval`, which does not exist — a user copying it hits a compile error. Corrected to `gdt::interval(entry.start, entry.end - 1)` **guarded by `entry.consumes_reference()`**, which is the important detail: unmapped / zero-ref-consuming reads have `end == start`, so the unguarded `end - 1` underflows `size_t`. Matches the `vcf_reader` usage example and the `sam_entry::end` field docstring.

**Label:** changed from `docs, refactor` to `docs` — there is no code restructuring here, only documentation.

CHANGELOG `### Changed` entry added under `[Unreleased]`.

PR #510, opened 2026-07-22
Add get_order() and index_names() to grove_view (fixes #509)

### LINKED ISSUE #509: grove_view read-side introspection parity: expose get_order() and index enumeration
## Gap

A feature scan of the in-memory `grove` against the partial reader `grove_view` found the query and graph-read paths at parity (`intersect`, `flanking`, `get_neighbors`, `get_edges`, `get_neighbors_if`, `get_edge_list`), but two **read-side introspection** accessors present on `grove` have no `grove_view` equivalent — even though the view already stores the underlying data.

### 1. `get_order()`

`grove::get_order()` returns the B+ tree order. `grove_view` **already reads and stores `order`** (`grove_view.hpp:269`, populated at `:305`) but never exposes it. Trivial getter.

### 2. Index enumeration

`grove::get_root_nodes()` lets a caller enumerate the index names (chromosomes). `grove_view` holds the same information in `index_roots` (`grove_view.hpp:273`) — it even iterates it internally for the all-index `intersect(query)` overload (`:107`) — but provides no public accessor. Without it, `intersect(query)` over all indices is opaque: a caller can't discover which indices are queryable.

`grove::get_root_nodes()` returns node pointers, which are meaningless for a lazy view. The view-appropriate analogue is a name list:

```cpp
[[nodiscard]] std::vector<std::string> index_names() const;   // unspecified order, like grove's unordered_map
```

## Proposed

Add to `grove_view`, alongside the existing `blocks_loaded()` / `block_count()` introspection getters:

```cpp
[[nodiscard]] int get_order() const;                          // mirrors grove::get_order()
[[nodiscard]] std::vector<std::string> index_names() const;   // analogue of grove::get_root_nodes()
```

Both back onto data the view already holds — no format change, no extra block loads, read path only.

## Out of scope (deliberately not added)

- **Mutation** (`insert*`, `add_edge`, `remove_*`, `compact`, `serialize`) — `grove_view` is a read-only reader by design.
- **Aggregate graph/key counts** (`edge_count`, `vertex_count`, `key_storage_size`, …) — the view only knows blocks paged in so far; a true total requires reading the whole file, contradicting the lazy design.
- **`grove_to_sif`** — a full SIF export pages in every block, defeating the partial reader.

## Downstream

Matching `pygenogrove` `GroveView` bindings can surface `get_order()` / `index_names()`; a docs update will follow once merged.

riasc:
### Added

A feature scan of the in-memory `grove` against the partial reader `grove_view` (issue #509) found the query and graph-read paths already at parity, but two **read-side introspection** accessors on `grove` had no view equivalent — despite the view already storing the underlying data. This adds both:

- **`int get_order() const`** — mirrors `grove::get_order()`. The view already reads and stores `order` at `open()`; this just exposes it.
- **`std::vector<std::string> get_index_names() const`** — the view-appropriate analogue of `grove::get_root_nodes()` (which returns node pointers, meaningless for a lazy view). Enumerates the index names from the in-memory directory. Without it, the all-index `intersect(query)` overload was opaque — no way to discover which indices are queryable. Returned in unspecified order, matching grove's unordered root map.

Both back onto data already held in memory from `open()` — no format change, no extra block loads, read path only.

### Out of scope (deliberately not added)

- **Mutation** (`insert*`, `add_edge`, `remove_*`, `compact`, `serialize`) — `grove_view` is read-only by design.
- **Aggregate graph/key counts** (`edge_count`, `vertex_count`, …) — the view only knows blocks paged in so far; a true total requires reading the whole file, contradicting laziness.
- **`grove_to_sif`** — a full SIF export pages in every block.

## Test plan

- [x] CI green across all platforms
- [x] Extended `GroveViewTest.MatchesEagerIntersectAcrossIndices` (order-4 chr1/chr2 fixture): asserts `view.get_order() == eager.get_order() == 4` and `get_index_names()` (sorted) `== {chr1, chr2}`.

## Downstream

Matching `pygenogrove` `GroveView` bindings can surface `get_order()` / `get_index_names()`; docs follow-up after merge.

## Summary by CodeRabbit

* **New Features**
  * Added read-only access to the configured B+ tree order for opened grove views.
  * Added the ability to list available index names from a grove view.
* **Tests**
  * Added coverage verifying reported tree order and index-name enumeration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **`grove_view` read-side introspection: `get_order()` and `get_index_names()`**: the partial (random-access) reader gains two accessors that were present on the in-memory `grove` but had no view equivalent, despite the view already holding the data from `open()`. `get_order()` returns the B+ tree order (mirrors `grove::get_order()`); `get_index_names()` returns `std::vector<std::string>` of every index (e.g. chromosome) in the `.gg`, in unspecified order — the view-appropriate analogue of `grove::get_root_nodes()` (which returns node pointers meaningless to a lazy view), letting a caller discover what `intersect(query)` / `flanking` can run against. Both read nothing extra (no block loads, no format change). Mutation, aggregate graph/key counts, and `grove_to_sif` were deliberately left out — the first is write-only, the latter two would force reading the whole file, contradicting the lazy design. ([#509](https://github.com/genogrove/genogrove/issues/509), [#510](https://github.com/genogrove/genogrove/pull/510))
+
 ## [0.25.4] - 2026-07-22
 
 ### Added
--- a/include/genogrove/structure/grove/grove_view.hpp
+++ b/include/genogrove/structure/grove/grove_view.hpp
@@ -247,6 +247,23 @@ class grove_view {
         return out;
     }
 
+    /// The B+ tree order the `.gg` was built with. Mirrors grove::get_order().
+    [[nodiscard]] int get_order() const { return order; }
+
+    /// Names of every index (e.g. chromosome) in the `.gg`, in unspecified order
+    /// (like grove's unordered root map). The view-appropriate analogue of
+    /// grove::get_root_nodes() — lets a caller discover what intersect(query) /
+    /// flanking can run against. Reads nothing extra: the directory is already
+    /// in memory from open().
+    [[nodiscard]] std::vector<std::string> index_names() const {
+        std::vector<std::string> names;
+        names.reserve(index_roots.size());
+        for (const auto& [name, root_id] : index_roots) {
+            names.push_back(name);
+        }
+        return names;
+    }
+
     /// Blocks paged in so far — for tests asserting a query is actually partial.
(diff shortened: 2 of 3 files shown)


riasc:
## Code Review

Reviewed against project conventions. All 7 CI jobs pass.

**No blocking findings.**

- **`get_order()`** — trivial `const` getter over the already-stored `order`, `[[nodiscard]]`, mirrors `grove::get_order()`. Correct.
- **`get_index_names()`** — copies names out of the in-memory `index_roots` directory with a `reserve`; no block loads, no format read. Returning `std::vector<std::string>` by value (rather than a `node*` map like `grove::get_root_nodes()`) is the right call for a lazy reader — there are no persistent root nodes to hand back, and `block_id` is an internal detail that shouldn't leak. Unspecified order is documented and matches grove's unordered root map.
- **Consistency**: both follow the `get_` getter convention (renamed from `index_names()` after review discussion) and the `[[nodiscard]]`-on-pure-getters policy.
- **Test**: extends the order-4 chr1/chr2 fixture — asserts `get_order()` parity with the eager grove (`== 4`) and `get_index_names()` (sorted) `== {chr1, chr2}`. Covers both new paths.

**Scope**: diff is exactly 2 files; the PR body explicitly documents what was left out (mutation, aggregate counts, SIF export) and why.

CHANGELOG `### Added` entry added under `[Unreleased]`; `feat` label correct.

PR #514, opened 2026-07-23
Bound block lengths without a per-block seek (fixes #513)

### LINKED ISSUE #513: Deserialization perf regression: per-block seek-to-end in require_backing_bytes (introduced by #511)
## Regression

`github-action-benchmark` flagged `BM_read_eager` on commit `61fc7da` (the #511 merge) as 2–7× slower than the previous benchmark (`6730416`):

| bench | current (61fc7da) | previous (6730416) | ratio |
|-------|-------------------|--------------------|-------|
| BM_read_eager/100/3   | 354.1 us  | 56.4 us   | 6.28 |
| BM_read_eager/100/32  | 47.9 us   | 22.7 us   | 2.11 |
| BM_read_eager/500/3   | 1870.8 us | 270.6 us  | 6.91 |
| BM_read_eager/1000/3  | 3782.4 us | 535.5 us  | 7.06 |
| BM_read_eager/5000/3  | 19070.4 us| 2672.3 us | 7.14 |
| BM_read_eager/10000/3 | 38351.8 us| 5357.0 us | 7.16 |

The slowdown scales with **block count**: order 3 (many small blocks) is ~7×, order 32 (fewer, larger blocks) is ~2×.

## Root cause

#511's `detail::require_backing_bytes(is, count, min, what)` bounds a file-controlled count by seeking to the stream end and back (`seekg(0, end)` + `seekg(cur, beg)`) to measure the remaining bytes. That is fine for the O(1) directory checks, but it was also placed in the **per-block loop** of `grove::deserialize` to bound each block's compressed length (`clen`):

```cpp
for (block b in num_blocks) {
    read clen;
    require_backing_bytes(is, clen, 1, "compressed block");  // <-- 2 file seeks EVERY block
    comp_buf.resize(clen); ...
}
```

`BM_read_eager` deserializes from an `std::ifstream`, so each block now costs two `lseek`s plus buffer-state disruption. With N blocks (N grows with dataset size, and is largest at low order), this dominates. `grove_view::read_block_raw` has the same per-load `require_backing_bytes(clen)` call (lazy, so it wasn't flagged, but it's the same waste).

## Fix

Keep the DoS bound, drop the per-block seeking:
- **Eager**: measure the remaining block-region bytes **once** before the loop and track a running `bytes_left` counter, decrementing by `sizeof(clen) + clen` each block. Reject `clen > bytes_left` with pure arithmetic — no per-block seek.
- **View**: store the file size once at scan time and validate `clen` against `file_size - block_offsets[b] - sizeof(clen)` — pure arithmetic, no per-block seek.

The O(1) directory checks (index count, index name length, block count) keep using the seek-based `require_backing_bytes` — they're called a handful of times and are not hot.

No behavior change to the bound itself (same values rejected); this is purely removing the per-block seek cost.

riasc:
### Fixed

`github-action-benchmark` flagged `BM_read_eager` as 2–7× slower after #511 (deserialization DoS bounds). Root cause: `detail::require_backing_bytes` measures the remaining stream bytes by seeking to the end and back, and #511 placed it in the **per-block loop** of `grove::deserialize` (and in `grove_view::read_block_raw`) to bound each block's compressed length `clen`. Reading from an `ifstream`, that is two `lseek`s plus buffer-state disruption per block — and block count scales with dataset size (worst at low order = many small blocks), giving the ~7× hit at order 3 and ~2× at order 32.

The fix keeps the exact same DoS bound but removes the per-block seek:

- **Eager** (`grove::deserialize`): measure the block-region byte budget **once** before the loop (`detail::remaining_bytes`, factored out of `require_backing_bytes`) and track it by subtraction — reject `clen > bytes_left` with arithmetic only.
- **View** (`grove_view`): store the file size once at scan time (`stream_size`) and bound `clen` against `stream_size - block_offsets[b] - sizeof(clen)` — no per-load seek.

The O(1) directory checks (index count, index name lengths, block count) still use the seek-based `require_backing_bytes` — they run a handful of times and aren't hot.

## Verification

- The DoS regression suite still fires: a huge `clen`, huge external-key count, etc. all throw `runtime_error` (the `HugeCompressedBlockLengthRejected` path now rejects via the arithmetic budget). A 1500-key grove round-trips.
- Local timing of the worst-regressed case (10000 keys, order 3, file deserialize, -O2): **~4.4 ms/iter**, back in the pre-#511 baseline band (~5.4 ms reported), versus the regressed ~38 ms.

No behavior change to the bound (same values rejected), no public API or `.gg` format change.

## Test plan

- [x] CI green across all platforms
- [x] Existing `SerializationDoSTest` (huge block/index/name/string counts, external-key cap, inflater cap) still asserts `runtime_error`; roundtrip suites confirm valid streams are unaffected. Perf itself is validated by the benchmark job on merge.

### DIFF
--- a/include/genogrove/structure/grove/grove_serialize.ipp
+++ b/include/genogrove/structure/grove/grove_serialize.ipp
@@ -376,6 +376,12 @@ public:
         detail::block_inflater inflater;
         std::string comp_buf;
         std::string raw_buf;
+        // Bytes remaining in the block region, measured once. Each block's
+        // file-controlled clen is validated against this by plain subtraction
+        // instead of a per-block seek-to-end — seeking every block dominated
+        // deserialize (#513). -1 means the source is not seekable, so the
+        // per-block clen bound is skipped (the truncated-read check still fires).
+        std::streamoff block_bytes_left = detail::remaining_bytes(is);
         try {
             // ---- read every block (length-prefixed, independently compressed) ----
             for (detail::block_id b = 0; b < num_blocks; ++b) {
@@ -390,10 +396,16 @@ public:
                 if (clen > static_cast<uint64_t>(std::numeric_limits<std::streamsize>::max())) {
                     throw std::runtime_error("Failed to deserialize grove: block length out of range");
                 }
-                // clen bytes must actually remain in the stream: reject a bogus
-                // multi-GB length before resize allocates it (OOM guard) rather
-                // than after the truncated read.
-                detail::require_backing_bytes(is, clen, 1, "compressed block");
+                // clen bytes must actually remain: reject a bogus multi-GB length
+                // before resize allocates it (OOM guard). Arithmetic only — the
+                // budget was measured once above.
+                if (block_bytes_left >= 0) {
+                    block_bytes_left -= static_cast<std::streamoff>(sizeof(clen));
+                    if (block_bytes_left < 0 || clen > static_cast<uint64_t>(block_bytes_left)) {
+                        throw std::runtime_error("Failed to deserialize grove: compressed block length exceeds remaining stream");
+                    }
+                    block_bytes_left -= static_cast<std::streamoff>(clen);
+                }
                 comp_buf.resize(static_cast<size_t>(clen));
                 is.read(comp_buf.data(), static_cast<std::streamsize>(clen));
                 if (is.gcount() != static_cast<std::streamsize>(clen)) {
--- a/include/genogrove/structure/grove/grove_view.hpp
+++ b/include/genogrove/structure/grove/grove_view.hpp
@@ -287,6 +287,7 @@ class grove_view {
     detail::block_id num_blocks = 0;
     detail::block_id ext_block_begin = 0;
     std::vector<std::uint64_t> block_offsets;  // block_id -> offset of [clen][bytes]
+    std::uint64_t stream_size = 0;  // file end; bounds each block's clen without a seek (#513)
     std::unordered_map<std::string, detail::block_id> index_roots;
 
     std::deque<key_t> key_storage;  // stable addresses for loaded keys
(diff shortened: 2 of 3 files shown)


riasc:
## Code Review

Reviewed against project conventions; ran the DoS harness and a timing check locally. All 7 CI jobs pass; no automated review comments (CodeRabbit rate-limited).

**No blocking findings.** The fix preserves the exact DoS bound while removing the per-block seek:

- **Eager**: `block_bytes_left` is measured once at the block-region start, then each iteration subtracts `sizeof(clen)` and `clen`. That matches exactly the bytes read from `is` per block (decompressed content is read from a separate `zis`), so the running budget is complete and correct. Non-seekable source → `-1` disables the bound (best-effort, unchanged), and the truncated-read check still fires.
- **View**: `stream_size` is captured once at scan end; `clen > stream_size - block_offsets[b] - sizeof(clen)` can't underflow because the scan already consumed block `b`'s `clen` field (so `block_offsets[b] + 8 <= stream_size`). Zero per-load seeks.
- **`pod_io`**: `remaining_bytes()` cleanly factored out; `require_backing_bytes` reuses it — behavior identical for the O(1) directory checks.

**Verification:** the `SerializationDoSTest` bounds still fire (huge `clen` now rejected via the arithmetic budget with an equivalent message); a 1500-key grove round-trips; local timing of the worst case (10000 keys / order 3, file deserialize, -O2) is ~4.4 ms/iter — back in the pre-#511 baseline band vs the regressed ~38 ms.

**CHANGELOG:** intentionally **no new entry**. This fixes a regression introduced by #511, which is still under `[Unreleased]` and has never shipped — a "fixed a regression you never saw" line would only confuse release-note readers. The #511 entry already describes the DoS bounds; this is an internal correction to that unreleased work.

**Label:** `bug` is correct (fixing a regression on `main`).

Issue #519, opened 2026-07-28
grove_remove_test builds its trees sorted, leaving find_leaf()'s routing thinly covered


riasc:
Follow-up from #518 (review finding, deliberately not addressed there).

## Problem

`find_leaf()` routes removal on each child's cached subtree maximum, so that removal descends into the same subtree insertion used. That descent changed in #518, but `grove_remove_test.cpp` barely exercises it:

- 10 of its 12 `insert_data` calls pass the `sorted` tag
- its only `shuffle` is of the **removal order**, not the insertion order

On a sorted-built tree the old and new routing agree by construction — a key greater than everything belongs in the rightmost child either way. That is exactly the blind spot that let #517 survive: every hot path inserts in sorted order, so the misroute was a no-op on all of them.

## What to do

Add unsorted-insert coverage to the removal suite: build trees with plain `insert()` / `insert_data()` over shuffled input at a small order (so the tree is several levels deep), then exercise the existing removal scenarios — underflow, borrow-from-left, borrow-from-right, merge, root collapse — and assert the validator invariants plus survivor reachability after each.

Worth doing as a pass over `grove_remove_test.cpp` rather than a single bolt-on case, since most of its scenarios would benefit from running against both a sorted-built and an unsorted-built tree.

## Not currently broken

Verified out-of-band while working on #518: order 4, 300 shuffled inserts, 100 removals — all 100 succeeded, 0 survivors lost, clean under ASan. This is a coverage gap, not a known defect.

Related: #517, #518

riasc:
## The residual unsorted-insert cost is mostly structural, not an implementation inefficiency

Investigated the 1.44-1.66x that #522 left behind at orders 3-15. Recording it here so it is not chased again as if it were a routing inefficiency.

### The two versions do not build the same tree

Same 5000 shuffled keys, same seed, node counts after insertion:

| order | v0.25.5 (misrouting) | current | |
|---|---|---|---|
| 3 | 5019 nodes, leaf fill 1.78/2 (89%) | **5787 nodes, fill 1.51/2 (75%)** | +15% nodes |
| 15 | 584 nodes, fill 9.63 | 554 nodes, fill 9.96 | slightly better |
| 32 | 245 nodes | 241 nodes | equal |
| 100 | 80 nodes, fill 63.3 | **68 nodes, fill 74.6** | -15% nodes |

[...]

Issue #517, opened 2026-07-26
Unsorted insert() misroutes keys in trees deeper than two levels — inserted key not findable by intersect()


riasc:
## Summary

Unsorted (out-of-order) `insert()` routes keys into the wrong child once the tree is deeper
than two levels. The key lands in a leaf too far to the right, which breaks the
globally-sorted leaf chain, and `intersect()` then cannot find it — including a query for the
key's own exact coordinates. `size()` counts it and graph edges reach it; only the spatial
index is wrong.

Sorted insertion (`insert_sorted`, `insert_data(..., sorted)`, bulk build) is unaffected: a
key that is greater than everything always belongs in the rightmost child, so the misroute is
a no-op there. That is why this has gone unnoticed — every hot path in the library inserts in
sorted order.

Found while investigating genogrove/pygenogrove#68 (reported there as a deserialization bug;
deserialization is not involved).

## Reproducer (plain `gdt::interval`, no data type, no I/O, no serialization)

```cpp
#include <genogrove/structure/all.hpp>
#include <genogrove/data_type/all.hpp>
#include <iostream>
#include <random>
#include <vector>

namespace ggs = genogrove::structure;
namespace gdt = genogrove::data_type;

int main() {
    for (int order : {3, 10, 100}) {
        for (int n : {200, 2000}) {
            ggs::grove<gdt::interval> g(order);
            std::vector<int> xs(n);
            for (int i = 0; i < n; ++i) xs[i] = i;
            std::shuffle(xs.begin(), xs.end(), std::mt19937(42));
            for (int i : xs) g.insert("chr1", gdt::key<gdt::interval>{gdt::interval(i*100, i*100+50)});

            // Walk the whole leaf chain via one all-covering query.
            const auto& ks = g.intersect(gdt::interval(0, 100000000), "chr1").get_keys();
            int inv = 0;
            for (size_t j = 0; j + 1 < ks.size(); ++j)
                if (ks[j]->get_value().get_start() > ks[j+1]->get_value().get_start()) ++inv;

            // Query every key with its own exact coordinates.
            int miss = 0;
            for (int i : xs)
                if (g.intersect(gdt::interval(i*100, i*100+50), "chr1").get_keys().empty()) ++miss;

            std::cout << "order=" << order << " n=" << n << " scanned=" << ks.size()
                      << " leaf-chain-inversions=" << inv
                      << " point-query-misses=" << miss << "\n";
        }
    }
}
```

Output on v0.25.5:

```
[...]

PR #522, opened 2026-07-28
Keep the routing cache off the node's size budget (perf regression from #518)

riasc:
## Summary

#518 fixed the routing bug in #517 but slowed every insert path — up to **1.54× on sorted insert** and **1.8× on unsorted**. The cause turned out to be none of the work that PR added.

It was `sizeof(node)`. The cached subtree maximum was stored as a `std::optional<key_type>`, taking the node from **80 to 104 bytes**. At small orders a grove is thousands of tiny nodes, so a 30% larger node means proportionally more cache misses across the entire workload.

The evidence for that diagnosis rather than an algorithmic one:

- the regression tracked **order inversely** — 1.8× at order 3, 1.04× at order 100 — i.e. it scaled with the node's *relative* size increase, not with any operation count
- removing the split-time separator narrowing changed 1.79× → 1.76×
- removing the per-level refresh changed 1.79× → 1.74×
- replacing the routing loop's child lookups with a contiguous scan over the parent's separators changed 1.80× → 1.88× (worse)

## Changes

1. **Cache a pointer to the maximum key** instead of a copy, the way nodes already reference keys.
2. **Pack `bool is_leaf` next to `int order`** so it sits in alignment padding instead of a tail byte. `sizeof(node<interval,int>)` is back to **exactly 80 bytes**.
3. **`insert_sorted` no longer does a map lookup per key** — it walks up from the leaf it already holds, re-looking-up only when a split actually moved the rightmost leaf.
4. **The per-level refresh is now conditional.** A subtree's maximum can only change if the inserted key exceeds it; a split reshuffles children but not which keys the subtree holds.
5. **`compact()` rebuilds the caches.** It rebuilds the grove's key storage, which leaves every cached pointer dangling — the one real bug the pointer representation reintroduces, and the reason #518 chose values.

## Results

Min of 5 interleaved rounds vs the v0.25.5 baseline (`5934b11`), orders 3-200, 1000 and 5000 keys:

| path | after #518 | this PR |
|---|---|---|
| sorted insert (all orders) | 1.30-1.54× | **1.03-1.10×** |
| bulk sorted / unsorted | 1.02-1.12× | **0.96-1.11×** |
| unsorted insert, order ≥ 32 | 1.04-1.38× | **0.95-1.02×** |
| unsorted insert, order 3-15 | 1.55-1.82× | 1.44-1.66× |

Interleaving the binaries mattered: measuring them back-to-back showed drift of 1.5×+ on code paths that were not touched at all.

### Still slow

Unsorted insert at orders 3-15 with large N. Routing costs one indirection more than the old separator comparison (child → cached pointer → key), and those trees run ~11 levels deep. Storing the maximum inline removes the hop but re-inflates the node, which measures worse. Left for the follow-up in #519; order 3 is a stress configuration, and real use runs order 100+, where this is at or below baseline.

### On the alert that started this

`BM_read_view`'s 2× is environmental — **1.00× measured locally**, and this PR touches no read path. Its CI ratio was flat at exactly 2.00× across a 100× range of N, which is a machine signature rather than an algorithmic one. For calibration, the docs-only commit `c8f98d9` swung that dashboard between 0.40× and 1.16×.

## Testing

`GroveCompactTest` already caught the dangling-pointer bug under ASan. Two new tests pin the properties this PR is about:

- `cache_does_not_grow_the_node` — asserts `sizeof(node)` against its expected field layout, so the cache cannot be re-inflated silently
- `compact_keeps_the_routing_cache_usable` — inserts *after* compacting, so routing must consult the rebuilt cache; fails visibly without a sanitizer

270/270 structure tests pass under ASan, and the #517 reproducer still reports 0 inversions and 0 misses.

## Test plan

- [x] `ctest` passes (structure, io, cli)
- [x] `sizeof(node)` unchanged vs v0.25.5 on CI compilers (gcc 13/14, clang 16/17/18, apple-clang 15/16) — `cache_does_not_grow_the_node` passed on all 7 jobs
- [ ] Benchmark run shows sorted/bulk paths back at baseline — **cannot be verified pre-merge**: the benchmark workflow runs on `push` to `main` only, never on PRs. Backed by local measurement until then
- [x] #517 routing invariants still hold (leaf-chain order, separator exactness, cached maxima)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved grove routing and insertion to correctly handle cached subtree maxima presence/updates.
  * Ensured `compact()` rebuilds routing maxima after key storage replacement so keys remain reachable.
  * Refined sorted insertion’s split/refresh flow to update routing caches only when required.
* **Tests**
  * Added routing-cache-focused tests covering compaction reachability and memory/layout expectations.
  * Updated tree invariant validation and node-move checks to reflect the new cached-maximum representation.


riasc:
## Code Review

Self-review. CI green on all 7 compiler jobs (910 tests, 100% passed); CodeRabbit posted a walkthrough with no findings. No bugs found in the diff, three things worth recording.

### Verified rather than assumed

The load-bearing claim of this PR is the conditional refresh: *"a split cannot change a subtree's maximum, only the inserted key can."* Worth stating why that holds, since it is not obvious — `split_internal_node` does promote a key into the parent. That promoted key is a **separator** (a bounding box whose value is then overwritten with an aggregate), and data keys live only in leaves. A split redistributes leaf keys between two leaves that remain in the same subtree, so the subtree's set of data keys is unchanged, and the cached pointer still names the same key object. Key storage is a `std::deque`, so addresses stay stable across `allocate_key`.

It is also well covered: the deep unsorted tests run at order 3 with hundreds of keys and constant splitting, and the validator asserts every node's cached maximum against an independently computed one. If the reasoning were wrong, those fail.

Also checked: move-constructor init order matches the new declaration order (no `-Wreorder`, confirmed on gcc 13/14 and clang 16/17/18); `refresh_subtree_max_upward` is not dead (the bulk-append path still uses it); no node is memcpy'd or copied by layout, and `serialize_block` writes fields explicitly, so reordering `is_leaf` is invisible to the format.

### Suggestion (optional): the sorted path duplicates an existing helper

`grove_insert.ipp` — the new upward walk in `insert_sorted`:

```cpp
for (auto* current = tail; current != nullptr; current = current->get_parent()) {
    current->refresh_subtree_max();
}
```

is exactly `refresh_subtree_max_upward(tail)`, which already exists and is still used by the bulk-append path. Calling the helper keeps the two sorted-append paths visibly identical.

### Note: the pointer cache is invisible to ASan when merely stale

Worth recording for whoever touches this next. The `compact()` bug was caught by ASan because that pointer became *freed*. The more likely future failure — a cached maximum that is stale but still points at a live key in the deque (`remove_key` unlinks keys from the tree but leaves them in storage) — produces no sanitizer report at all. It silently misroutes, exactly like #517.

That makes `tree_validator`'s cached-maximum assertion the only thing standing between a missed refresh and a silent corruption. Not an argument against the pointer representation (the value-based one was equally silent when stale), just a reason not to weaken that assertion later.

### Process: this PR's central claim cannot be verified by CI

`.github/workflows` runs the benchmark job on `push` to `main` only, never on pull requests. So "sorted and bulk paths are back at baseline" — the entire point of this PR — is backed only by local measurement until after it merges. The QC item for it stays unchecked; if the numbers do not reproduce on the CI machine, we find out post-merge from the next alert.


---

## Update — findings addressed

| Finding | Status | Where |
|---|---|---|
| `insert_sorted` duplicates `refresh_subtree_max_upward` | **Addressed** | 5fcb1ae — both sorted-append paths now call the helper. 270/270 tests pass under ASan; sorted insert stays at 1.03-1.10x of baseline across orders 3-200, so the dedup is behaviour- and performance-neutral |
| Stale pointer cache is invisible to ASan | **Not addressed** | Nothing to fix — it is a property of the representation, recorded so the validator's cached-maximum assertion is not weakened later |
| Benchmarks do not run on PRs | **Not addressed** | Out of scope here. The QC item is left unchecked with the reason inline; the real numbers arrive with the first post-merge alert |

PR #527, opened 2026-07-30
Randomized removal coverage over unsorted-built trees

### LINKED ISSUE #517: Unsorted insert() misroutes keys in trees deeper than two levels — inserted key not findable by intersect()

riasc:
## Summary

Adds a randomized removal case to `grove_remove_test.cpp`: build a tree **unsorted** and remove all 200 keys in random order at orders 3, 4 and 6, asserting on every removal that the key is still findable and validating the tree as it drains.

**This does not close a demonstrated gap**, and the framing in #519 (and in earlier revisions of this PR) overstated the case. Two rounds of mutation testing established what it actually does, and both results are worth having on record.

### What it does not do: pin `find_leaf`'s descent

#519 argued the suite could not distinguish correct from incorrect routing in `find_leaf`. Reverting `find_leaf` to the pre-#518 separator comparison and re-running leaves **all 40 removal tests passing**, this one included.

`find_leaf` walks the leaf chain forward after descending, and the old scan cannot overshoot: the key being removed lies in its child's subtree, so it necessarily overlaps that child's bounding box, and the `!overlaps(...)` escape halts the scan there at the latest. Landing early is recovered by walking right. So with correct caches the two routings are indistinguishable through the public API, and aligning them in #518 was defensive consistency rather than a fix. Detail in [this comment](https://github.com/genogrove/genogrove/issues/519#issuecomment-5133871481).

### What it does do: check the cache is good enough to route by, on every removal

Deleting the cache refresh from each rebalance path in a scratch copy and re-running:

| refresh deleted from | this test | rest of the suite |
|---|---|---|
| `merge_with_sibling` | fails | 7 tests fail |
| `try_borrow_from_right` | fails | 14 tests fail |
| `try_borrow_from_left` | fails | 10 tests fail |

So it catches nothing the suite misses today. But the *way* it fails is informative — `remove_key` returns false for a key still in the tree:

```
remove_key could not find the key at 840 after 22 removals — the descent reached the wrong subtree
```

That exposes an asymmetry introduced by #518. The old separator scan could only undershoot, which the forward chain walk recovers. Max-based routing advances while `key > child_max`, so a stale maximum that is too small makes it advance *past* the correct child — and the chain walk only goes right, so there is no recovering. Routing on real maxima is what fixes #517, but it is cache-dependent in a way the old comparison was not, which makes the validator's cached-maximum assertions load-bearing.

The per-removal assertion here is that check, run 200 times per order against shapes the hand-built sorted cases never produce.

### Verdict

Keep it for randomized shape diversity and the per-removal cache check, not as a gap-closer. On that basis #519 is better **closed** than implemented — its premise does not hold.

## Test plan

- [x] `ctest` passes on all platforms — 911 tests, 100% passed on all 7 jobs
- [x] The new case passes at orders 3, 4 and 6
- [x] No existing removal test changed behaviour — additive only; the 40-test suite passes unchanged
- [x] The test is capable of failing — verified by deleting each rebalance path's cache refresh in a scratch copy

## Summary by CodeRabbit

* **Tests**
  * Added randomized removal coverage across multiple tree orders.
  * Validates key removal, tree integrity, and remaining key reachability as the tree is emptied.
  * Confirms indexes are fully cleaned up after all removals.

* **Documentation**
  * Documented the expanded randomized removal test coverage in the unreleased changelog.


riasc:
## Review of `test/519-unsorted-removal` (commit `5ac0caa`)

### Summary

Single commit adding one test: `RandomisedRemovalFromUnsortedTree` in `tests/structure/grove_remove_test.cpp`. No production code changes.

### What it tests

Builds a B+ tree via **shuffled (unsorted) insertion** — the insert path that routes on subtree maxima (#518), producing tree shapes the sorted-append path never creates. Then removes all keys in random order at orders 3, 4, 6, validating structure invariants and survivor reachability every 25 removals.

This exercises the borrow/merge/collapse rebalance machinery against irregular tree topologies. The comment block (L931–948) is unusually thorough about what it covers and, importantly, what it *doesn't*.

### Strengths

1. **Multiple orders** (3, 4, 6) — different `min_keys` thresholds, different internal branching factors.
2. **Deterministic seeds** derived from order — reproducible across runs.
3. **Validation frequency** (every 25 removals) is a pragmatic tradeoff. The `ASSERT_TRUE` on `remove_key` (L966) provides a hard stop on misrouting — a stranded key fails the `find_leaf` descent immediately.
4. **Survivor reachability check** (L982–985) — verifies every surviving key is still discoverable via `intersect`, not just structurally present.
5. **Index cleanup** verified — `root_nodes` erased and `indexed_vertex_count == 0` after final removal.
6. **No library changes** — purely additive test coverage.

### Issues found

**Minor**

1. **Missing blank line** between the closing `}` of `RoundtripSerializationAfterCompact` (L926) and the section comment (L927). Formatting only.

2. **Validation gap in the last ~25 removals**: If the `step % 25 == 0` guard doesn't fire for the final batch (200 removals: validations at steps 0, 25, 50, 75, 100, 125, 150, 175 — the last ~24 removals are unvalidated). Mitigated by:
   - The `ASSERT_TRUE` on each `remove_key` — corrupted routing surfaces immediately as a failed `find_leaf`.
   - The final assertion that the index is fully cleaned up.

   Worth calling out but not a blocker.

3. **Comment-as-design-doc** (L931–948): This reads like a commit message inlined as a code comment. The blind-spot documentation is valuable, but the same info lives in the commit message already.

**Not issues (anticipated but checked):**

- No coverage of compaction + unsorted removal — the compact path is algorithmically independent of build order, so fine.
- No graph edge testing in unsorted context — `remove_key` already calls `graph_data.remove_all_edges`, and graph overlay has no dependence on tree topology.

### Verdict

**Approved.** Clean, well-scoped, correctly designed. The test adds meaningful coverage of a path that previously had none.

riasc:
## Code Review

Self-review. CI green on all 8 checks; CodeRabbit posted no findings. Two things in my own test, one of which undercuts part of its purpose.

### Bug: the validation cadence skips the endgame, where the most interesting rebalancing happens

```cpp
if (remaining > 0 && step % 25 == 0) { ...validate... }
```

That validates at steps 0, 25, … 175 — so the last 25 removals run entirely unvalidated. Those are exactly the ones worth watching: as a tree drains, merges cascade, the tree loses levels, and `collapse_root` fires. Right now the only thing checked across that stretch is that each `remove_key` returned true and that the index eventually disappears, so a separator or cached-maximum left stale during a cascading merge or a root collapse would pass unnoticed.

Given the test exists to exercise rebalance paths against unsorted-built shapes, leaving out the phase with the densest rebalancing is the wrong 12% to skip. Validating whenever few keys remain, in addition to the periodic check, costs nothing at these sizes — the whole test runs in 9 ms.

### Suggestion (optional): spelling is inconsistent with the codebase

`RandomisedRemovalFromUnsortedTree`, and "randomised" in the comment. The codebase is overwhelmingly `-ize`: 174 occurrences of `serialize`/`normalize`/`initialize`/`optimize` against 4 `-ise` forms. Test names surface in CI logs and `--gtest_filter` arguments, so consistency has some practical value. `RandomizedRemovalFromUnsortedTree`.

### Note: where reachability actually comes from in this file (no action)

Worth recording, because the same name means two different things depending on the file. `validate_grove_index` in `grove_remove_test.cpp` takes a **root node** and wraps `validate_tree_structure`; the one in `key_type_grove_test.hpp` takes a **grove** and additionally queries every leaf key through `intersect()`. They coexist as overloads rather than clashing, but only the second checks reachability.

So in this test, reachability is covered by the survivor-count query against an all-covering interval, not by the validator call. That is sufficient — a stranded key makes the count come up short — but anyone reading the validator call and assuming it covers reachability would be wrong.

### Checked and fine

- Each `remove_key` is asserted to return true, which is the strongest per-step assertion available: removal is by pointer identity, so a true return means the descent reached the leaf actually holding that key.
- The all-covering query spans `[0, 2000]` against a maximum interval of `[1990, 1995]`, so no survivor can fall outside it.
- Keys are returned in coordinate order while insertion is shuffled, so the removal order is independent of the insertion order rather than accidentally correlated.
- Three orders (3, 4, 6) all produce multi-level trees at 200 keys, so borrow and merge fire in every configuration rather than only the deepest.

